PR #93, opened 2017-09-20
Contract cleanup, verification

### LINKED ISSUE #91: uRaiden contract - dont't optimize
optimize to false in populus.json

loredanacirstea:
closes #91 

- Removed commented code
- Print ABI encoded arguments after deployment
- redeployed in Kovan and verified the `RaidenMicroTransferChannels` contract


contracts/deploy/sign.py
```diff
@@ -59,7 +66,8 @@ def format_int(value, size):
             msg += arg
         elif isinstance(arg, str):
             if is_0x_prefixed(arg):
-                msg += decode_hex(arg)
+                arg = pad_left(remove_0x_prefix(arg), 64, '0')
```
jannikluhn: I don't think that's correct, as there are `byte8`'s etc. in Solidity which are not padded. I would just leave it like it is (the caller can pad if he needs to).

contracts/deploy/sign.py
```diff
@@ -59,7 +66,8 @@ def format_int(value, size):
...
+                msg += decode_hex(add_0x_prefix(arg))
```
jannikluhn: adding the prefix again isn't necessary, `decode_hex` handles strings with and without prefixes

contracts/contracts/RaidenMicroTransferChannels.sol
```diff
@@ -128,7 +128,13 @@ contract RaidenMicroTransferChannels {
         return str;
     }
 
-    // 56014 gas cost
+    /*
+     *  Temporary implementation of verifyBalanceProof.
+     *  Message string reproduction is a workaround
+     *  until https://github.com/ethereum/EIPs/pull/712 is implemented.
```
jannikluhn: Just a note: That EIP hasn't been accepted yet, so it's not certain that it will ever be implemented.

PR #101, opened 2017-09-20
Support JSON-encoded private keys, as generated by geth/parity

### LINKED ISSUE #100: Support JSON-encoded private keys from Geth/Parity
Today, we expect the privatekey to be a 32 bytes hex-encoded sequence, which is very unusual to find in the real world. We must support Geth and Parity's [keystore](https://theethereum.wiki/w/index.php/Accounts,_Addresses,_Public_And_Private_Keys,_And_Tokens#UTC_JSON_Keystore_File) format directly, asking the user for password, if needed.

andrevmatos:
Support JSON private keys.
Old format (raw hex encoded) is still supported, but will generate a warning, and should be considered deprecated in the future, as it's less secure and not standardized.

It will **interactively** ask for the password for decryption of the private key.
Optionally, CLI tools now have an option to pass the path of a file containing the password, for non-interactive opening of the private key (`docker` environments, scripting): `--private-key-password-file` for the proxy, `--key-password-path` for python Client/m2m.

Fix #100 .

### DIFF
--- a/microraiden/microraiden/close_all_channels.py
+++ b/microraiden/microraiden/close_all_channels.py
@@ -30,6 +30,7 @@ log = logging.getLogger('close_all_channels')
 @click.command()
 @click.option(
     '--private-key',
+    required=True,
     help='Path to private key file of the proxy',
     type=click.Path(exists=True, dir_okay=False, resolve_path=True)
 )
--- a/microraiden/microraiden/close_all_channels.py
+++ b/microraiden/microraiden/close_all_channels.py
@@ -8,7 +8,6 @@ import traceback
 import click
 from eth_utils import (
     decode_hex,
-    is_hex,
     is_same_address,
 )
 from ethereum.tester import TransactionFailed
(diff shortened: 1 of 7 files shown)


microraiden/microraiden/close_all_channels.py
```diff
@@ -44,18 +49,9 @@
     default=None,
     help='Ethereum address of the channel manager contract'
 )
-def main(private_key, state_file, channel_manager_address):
+def main(private_key, private_key_password_file, state_file, channel_manager_address):
+    private_key = utils.get_private_key(private_key, private_key_password_file)
     if private_key is None:
```
pcppcp: Can you make `--private-key` click option mandatory? It doesn't really make sense to start the proxy without privkey and click will display a nice error message if the option is missing.

microraiden/microraiden/utils.py
```diff
@@ -14,3 +26,34 @@ def check_permission_safety(path):
     """
     f_stats = os.stat(path)
     return (f_stats.st_mode & (stat.S_IRWXG | stat.S_IRWXO)) == 0 and f_stats.st_uid == os.getuid()
+
+
+def get_private_key(key_path, password_path=None):
+    if key_path is None or not os.path.exists(key_path):
```
pcppcp: From usability perspective, I think it's better to split this into

```
assert key_path is not None  #  probably developer's mistake
if not os.path.exists(key_path):  
   log.fatal("%s: no such file")  # user's mistake 
```

microraiden/microraiden/close_all_channels.py
andrevmatos: Done.  `click.option` has a `required` keyword, although options usually are meant to be... optional. But should work.

PR #103, opened 2017-09-21
Travis ci

pcppcp:
this PR enables Travis builds and also reduces test execution time on tester chain

### DIFF
--- a/microraiden/microraiden/test/fixtures/channel_manager.py
+++ b/microraiden/microraiden/test/fixtures/channel_manager.py
@@ -3,11 +3,6 @@ import types
 from microraiden.channel_manager import ChannelManager, Blockchain
 
 
-@pytest.fixture
-def channel_managers_count():
-    return 2
-
-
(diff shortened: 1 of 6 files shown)


microraiden/microraiden/test/fixtures/channel_manager.py
```diff
@@ -39,8 +34,12 @@ def fail(greenlet):
 @pytest.fixture
 def channel_manager(web3, receiver_privkey, make_channel_manager_proxy, token_contract, use_tester,
                     mine_sync_event):
+    if use_tester:
+        snapshot_data = web3.testing.snapshot()
```
jannikluhn: would rename `snapshot_data` to `snapshot_id` (it's not the actual snapshot, but only a reference)

pcppcp: Yes you are right, I've got it mistaken for another function that returns binary-serialized block as a snapshot.

PR #117, opened 2017-09-26
Check contract existence

jannikluhn:
Checks that the code at the channel manager address is not empty (e.g. if the address is wrongly configured).

A better solution would be to actually compare the bytecode against the one stored in the contracts metadata file. However, this doesn't work right now as the code differs from chain to chain (due to differing token addresses) and we keep only one variant (presumably the one that has been deployed last).


microraiden/microraiden/make_helpers.py
```diff
@@ -19,15 +17,22 @@
...
 def make_contract_proxy(web3, private_key, contract_address):
-    contracts_abi_path = os.path.join(os.path.dirname(__file__), 'data/contracts.json')
-    abi = json.load(open(contracts_abi_path))['RaidenMicroTransferChannels']['abi']
-    return ChannelContractProxy(web3, private_key, contract_address, abi, config.GAS_PRICE,
-                                config.GAS_LIMIT)
+    metadata = config.CONTRACT_METADATA['RaidenMicroTransferChannels']
+    bytecode = web3.eth.getCode(contract_address)
+    if bytecode == '0x':
+        raise ValueError('No contract deployed at address {}'.format(contract_address))
+    return ChannelContractProxy(
```
pcppcp: do `web3.eth.getCode` inside of the `ChannelContractProxy` class and raise an exception if check fails

pcppcp:
fixed by another PR

Issue #126, opened 2017-11-05
filename should not be part of ChannelManagerState


nkonovalov:
I can't move the state file otherwise to a different path.

loredanacirstea:
Fixed in https://github.com/raiden-network/microraiden/pull/145

pcppcp:
btw, channel manager still references the state file.

See 
`channel_manager.py:295-296` (this raises an exception when starting uRaiden on demo site w/o synced chain)
https://github.com/jannikluhn/microraiden/blob/5f69f69a7be82dc9ee87b2c7f1d788ba65b23260/microraiden/microraiden/channel_manager.py#L295
and 
`channel_manager.py:257`
https://github.com/jannikluhn/microraiden/blob/5f69f69a7be82dc9ee87b2c7f1d788ba65b23260/microraiden/microraiden/channel_manager.py#L257 
 (there should be key check before calling `pop()`)

pcppcp:
Fixed in PR #180

PR #145, opened 2017-11-09
Drop filename and tmp filename from state file

jannikluhn:
see #126


freehere107:
but microraiden/microraiden/channel_manager.py:299 will raise 
KeyError: 'tmp_filename' and KeyError: 'filename'

loredanacirstea:
@freehere107 , is the issue solved if you remove the cached state files and restart microraiden?

You will see something like `INFO:filelock:Lock 4581394696 acquired on /Users/loredana/Library/Application Support/microraiden/0xffa52825_0x5601ea84.json` when starting microraiden. Cached state files are those `json` files under the `microraiden` folder. Depending on your OS, the files might be in another location.

PR #155, opened 2017-11-20
Solc 0.4.17 fixes

LefterisJP:
Still leaves one warning though.

```
contracts/contracts/RaidenMicroTransferChannels.sol:167:26: Warning: Function declared as view, but this expression (potentially) modifies the state and thus requires non-payable (the default) or payable.                                                                    
        address signer = ECVerify.ecverify(prefixed_message_hash, _balance_msg_sig);                                                                                                                                                                                            
                         ^--------------------------------------------------------^
```

We can't mark the ECRecover lib as pure due to it's assembly code doing a `CALL` into what is I assume the ecrecover precompiled contract, but the solidity compiler is not smart enough to recognize this yet so if you mark all the contracts in ECRecover lib as pure as they should be [this](https://github.com/raiden-network/microraiden/blob/master/contracts/contracts/lib/ECVerify.sol#L5) one fails with:

```
contracts/contracts/lib/ECVerify.sol:16:20: Error: Function declared as pure, but this expression (potentially) modifies the state and thus requires non-payable (the default) or payable.                                                                                      
            ret := call(3000, 1, 0, size, 128, size, 32)              
```

### DIFF
--- a/contracts/contracts/RaidenMicroTransferChannels.sol
+++ b/contracts/contracts/RaidenMicroTransferChannels.sol
@@ -76,7 +76,7 @@ contract RaidenMicroTransferChannels {
     /// @dev Constructor for creating the Raiden microtransfer channels contract.
     /// @param _token The address of the Token used by the channels.
     /// @param _challenge_period A fixed number of blocks representing the challenge period after a sender requests the closing of the channel without the receiver's signature.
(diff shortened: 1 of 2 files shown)


LefterisJP:
I guess should merge after #122 @loredanacirstea

LefterisJP:
@loredanacirstea rebased

PR #156, opened 2017-11-20
Add versioning to the RaidenMicroTransferChannels contract

### LINKED ISSUE #99: Contract versioning
We should support contract versioning - if the contract is obsolete, it should i.e. have a method that returns new contract address.

(personally, I think it'd be better to just use contract address sent in http headers and have a 'obsolete' flag in the contract that'll abort all public methods if set)

loredanacirstea:
closes #99 

- string variable for compatibility with semantic versioning
- setter & getter for latest contract version Ethereum address
- setter only callable by owner
- unit test

### DIFF
--- a/contracts/contracts/RaidenMicroTransferChannels.sol
+++ b/contracts/contracts/RaidenMicroTransferChannels.sol
@@ -13,6 +13,8 @@ contract RaidenMicroTransferChannels {
     address public owner;
     address public token_address;
     uint8 public challenge_period;
+    string public constant version = '1.0.0';
+    address public latest_version_address;
     string constant prefix = "\x19Ethereum Signed Message:\n";
 
     Token token;
@@ -37,6 +39,11 @@ contract RaidenMicroTransferChannels {
 
     modifier isToken() {
         require(msg.sender == token_address);
+        _;
+    }
+
+    modifier isOwner() {
+        require(msg.sender == owner);
         _;
     }
 
@@ -88,9 +95,21 @@ contract RaidenMicroTransferChannels {
         challenge_period = _challenge_period;
     }
 
+    /// @dev Sets the address for the latest contract version
+    /// @param _latest_version_address The address for the latest contract version.
+    function setLatestVersionAddress(address _latest_version_address) public isOwner {
+        require(addressHasCode(_latest_version_address));
+        latest_version_address = _latest_version_address;
+    }
+
     /*
      *  Public helper functions (constant)
      */
+
+    /// @dev Returns the address for the latest contract version.
+    function getLatestVersionAddress() public view returns(address) {
+        return latest_version_address;
+    }
 
     /// @dev Returns the unique channel identifier used in the contract.
     /// @param _sender The address that sends tokens.
--- a/contracts/contracts/RaidenMicroTransferChannels.sol
+++ b/contracts/contracts/RaidenMicroTransferChannels.sol
@@ -105,12 +105,6 @@ contract RaidenMicroTransferChannels {
     /*
      *  Public helper functions (constant)
      */
-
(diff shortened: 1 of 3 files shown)


contracts/contracts/RaidenMicroTransferChannels.sol
```diff
@@ -88,10 +95,22 @@ contract RaidenMicroTransferChannels {
         challenge_period = _challenge_period;
     }
 
+    /// @dev Sets the address for the latest contract version
```
LefterisJP: Hmm I am a bit confused. Why add a version setter? Once the code of the contract changes the version should also change. At least the `PATCH` part from `MAJOR.MINOR.PATCH`.

contracts/contracts/RaidenMicroTransferChannels.sol
```diff
@@ -88,10 +95,22 @@ contract RaidenMicroTransferChannels {
...
+        require(addressHasCode(_latest_version_address));
+        latest_version_address = _latest_version_address;
+    }
+
     /*
      *  Public helper functions (constant)
      */
 
+    /// @dev Returns the address for the latest contract version.
+    function getLatestVersionAddress() public view returns(address) {
```
LefterisJP: Why a getter? It should be constant. Similar to my comment above.

contracts/contracts/RaidenMicroTransferChannels.sol
loredanacirstea: In case someone finds this contract and wants to use it and is not sure if it is the last version, it's nice to be able to check this and have a way to get the last contract version deployed.
Of course, this means taking care of updating all deployed versions when deploying a new one - so the setter will be used in the deployment process.

This was the reasoning. More details in the comments here https://github.com/raiden-network/microraiden/issues/99

contracts/contracts/RaidenMicroTransferChannels.sol
loredanacirstea: It gets the address for the last version contract, not the version itself.

contracts/contracts/RaidenMicroTransferChannels.sol
```diff
@@ -88,10 +95,22 @@ contract RaidenMicroTransferChannels {
         challenge_period = _challenge_period;
     }
 
+    /// @dev Sets the address for the latest contract version
+    /// @param _latest_version_address The address for the latest contract version.
+    function setLatestVersionAddress(address _latest_version_address) public isOwner {
+        require(addressHasCode(_latest_version_address));
```
ice09: not sure if this check has to be here, I could use a precalculated address here, which I want to deploy the contract to _after_ I changed all "references"

loredanacirstea: I didn't understand how this works, especially the "after I changed all "references"".
So, you precalculate the new contract address and use it for deployment. Then you still have to change `_latest_version_address` for the old contracts with this setter.

ice09: Yes, that's correct, doesn't make sense this way. I thought of a scenario where it's crucial to set the address before I deploy, but there is no such scenario.

contracts/contracts/RaidenMicroTransferChannels.sol
loredanacirstea: @LefterisJP , I will assume a misunderstanding (version vs. address) and will merge this PR for now. If not, please open a issue.
Removed the redundant getter though.

LefterisJP: Hmm I don;t really think this is required but if you would like to have it then you can have it. I rather think just having an ENS record pointing to the latest contract should be sufficient

PR #157, opened 2017-11-21
Contract ecverify

loredanacirstea:
- remove safer_ecrecover (old code from August 2016, Solidity's ecrecover has been improved since then)
- add require for input checks
- check for ecrecover returning 0 on error


contracts/contracts/lib/ECVerify.sol
```diff
@@ -2,51 +2,36 @@ pragma solidity ^0.4.17;
...
+            v += 27;
+        }
 
-        if (v != 27 && v != 28)
-            return (false, 0);
+        require(v == 27 || v == 28);
 
-        return safer_ecrecover(hash, v, r, s);
-    }
+        var addr = ecrecover(hash, v, r, s);
```
ice09: As discussed, please use address as type for addr, should work as well.

contracts/contracts/lib/ECVerify.sol
```diff
@@ -2,51 +2,36 @@ pragma solidity ^0.4.17;
 
 library ECVerify {
 
-    function safer_ecrecover(bytes32 hash, uint8 v, bytes32 r, bytes32 s) internal returns (bool, address) {
-        bool ret;
-        address addr;
+    function ecverify(bytes32 hash, bytes signature) internal pure returns (address) {
```
ice09: As discussed, please use `(address addr)` as return type for addr, should work as well.

contracts/contracts/lib/ECVerify.sol
loredanacirstea: Yes, was not sure it would work if ecrecover returned 0 on error.

For documenting this, ecrecover returns `u256(0)` on error: https://github.com/ethereum/solidity/pull/838/files#diff-ee617ea709ab57f19fa067e365a9b4f8R1456

Issue #134, opened 2017-11-09
Internal audit of the smart contract


loredanacirstea:
# Internal audit Part 1

21st November
Participants: @loredanacirstea, @czepluch, @ice09 

**Scope:**
- https://github.com/raiden-network/microraiden/blob/master/contracts/contracts/RaidenMicroTransferChannels.sol
- Contract versioning PR https://github.com/raiden-network/microraiden/pull/156
- ECVerify refactor PR https://github.com/raiden-network/microraiden/pull/157

**Agenda**: go through the contracts code and PRs for an overview and discussions

**Outcome:**
1. improve `settleChannel` logic, get rid of `min()` & `max()` -> fixed https://github.com/raiden-network/microraiden/pull/163 
2. change `close` to `cooperativeClose` & `uncooperativeClose` -> fixed https://github.com/raiden-network/microraiden/pull/163 
3. code_style & comment fixes -> fixed https://github.com/raiden-network/microraiden/pull/163 
4. ECVerify.sol - use `address signature_address` instead of `var addr` -> fixed in https://github.com/raiden-network/microraiden/commit/be16aa245eb5766609ecd588a6988b9b8225896b

5. `_sender` to `_sender_address` & similar (not changing this for now in events because -> python code refactor; leaving for future release); -> fixed https://github.com/raiden-network/microraiden/pull/163 
6. possible issues with `uint192(_deposit)` - `_deposit` is `uint256`
https://github.com/raiden-network/microraiden/blob/1f175393c604a7b682374110690d25790a7e7808/contracts/contracts/RaidenMicroTransferChannels.sol#L217 -> fixed in https://github.com/raiden-network/microraiden/pull/204

7. document `memcpy`, `concat` - taken from https://github.com/Arachnid/solidity-stringutils/blob/master/strings.sol for temporary use until we can implement `eth_signTypedData`
fixed (removed it) https://github.com/raiden-network/microraiden/pull/176

LefterisJP:
For the versioning PR I left some comments in a new issue along with a proposal: https://github.com/raiden-network/microraiden/issues/162

berndbohmeier:
## Internal Audit Part 2
Participants: Bernd Bohmeier

#### scope:
https://github.com/raiden-network/microraiden/blob/master/contracts/contracts/RaidenMicroTransferChannels.sol

[...]

PR #163, opened 2017-11-21
Contract internal audit 1

loredanacirstea:
As documented in https://github.com/raiden-network/microraiden/issues/134

- Better variable naming - contract
- Fix comments
- Rename `close` functions to `cooperativeClose` & `uncooperativeClose`
- Refactor `settleChannel`, remove `min()` and `max()`
- Add overflow check
- Replace `!= 0` with `>0` for uint, code style fix

### DIFF
--- a/contracts/contracts/RaidenMicroTransferChannels.sol
+++ b/contracts/contracts/RaidenMicroTransferChannels.sol
@@ -31,8 +31,11 @@ contract RaidenMicroTransferChannels {
 
     // 28 (deposit) + 4 (block no settlement)
     struct Channel {
-        uint192 deposit; // mAX 2^192 == 2^6 * 2^18
-        uint32 open_block_number; // UNIQUE for participants to prevent replay of messages in later channels
+        // mAX 2^192 == 2^6 * 2^18
+        uint192 deposit;
+
+        // UNIQUE for participants to prevent replay of messages in later channels
+        uint32 open_block_number;
     }
 
     struct ClosingRequest {
@@ -78,10 +81,7 @@ contract RaidenMicroTransferChannels {
         address indexed _receiver,
         uint32 indexed _open_block_number,
         uint192 _balance);
-    event GasCost(
-        string _function_name,
-        uint _gaslimit,
-        uint _gas_remaining);
+
 
     /*
      *  Constructor
@@ -89,8 +89,8 @@ contract RaidenMicroTransferChannels {
 
     /// @dev Constructor for creating the uRaiden microtransfer channels contract.
     /// @param _token_address The address of the Token used by the uRaiden contract.
-    /// @param _challenge_period A fixed number of blocks representing the challenge
-    /// period after a sender requests the closing of the channel without the receiver's signature.
+    /// @param _challenge_period A fixed number of blocks representing the challenge period
+    /// after a sender requests the closing of the channel without the receiver's signature.
     function RaidenMicroTransferChannels(address _token_address, uint8 _challenge_period) public {
         require(_token_address != 0x0);
         require(addressHasCode(_token_address));
@@ -116,7 +116,8 @@ contract RaidenMicroTransferChannels {
     /// @dev Returns the unique channel identifier used in the contract.
     /// @param _sender The address that sends tokens.
     /// @param _receiver The address that receives tokens.
-    /// @param _open_block_number The block number at which a channel between the sender and receiver was created.
+    /// @param _open_block_number The block number at which a channel between the
+    /// sender and receiver was created.
     /// @return Unique channel identifier.
     function getKey(
         address _sender,
@@ -131,7 +132,8 @@ contract RaidenMicroTransferChannels {
 
     /// @dev Returns a hash of the balance message needed to be signed by the sender.
     /// @param _receiver The address that receives tokens.
-    /// @param _open_block_number The block number at which a channel between the sender and receiver was created.
+    /// @param _open_block_number The block number at which a channel between the
+    /// sender and receiver was created.
     /// @param _balance The amount of tokens owed by the sender to the receiver.
     /// @return Hash of the balance message.
     function getBalanceMessage(
@@ -154,14 +156,16 @@ contract RaidenMicroTransferChannels {
      *  Temporary implementation of verifyBalanceProof.
      *  Message string reproduction is a workaround
      *  until https://github.com/ethereum/EIPs/pull/712 is implemented.
-     *  Reason: offer the _sender security that the Dapp sends the same balance as in the signed message
+     *  Reason: offer the _sender security that the Dapp sends the same balance as in
+     *  the signed message
      */
 
     /// @dev Returns the sender address extracted from the balance proof.
     /// @param _receiver The address that receives tokens.
-    /// @param _open_block_number The block number at which a channel between the sender and receiver was created.
+    /// @param _open_block_number The block number at which a channel between the
+    /// sender and receiver was created.
     /// @param _balance The amount of tokens owed by the sender to the receiver.
-    /// @param _balance_msg_sig The balance message signed by the sender or receiver.
+    /// @param _balance_msg_sig The balance message signed by the sender.
     /// @return Address of the balance proof signer.
     function verifyBalanceProof(
         address _receiver,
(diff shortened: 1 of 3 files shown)


contracts/contracts/RaidenMicroTransferChannels.sol
```diff
@@ -31,8 +31,11 @@ contract RaidenMicroTransferChannels {
 
     // 28 (deposit) + 4 (block no settlement)
     struct Channel {
-        uint192 deposit; // mAX 2^192 == 2^6 * 2^18
-        uint32 open_block_number; // UNIQUE for participants to prevent replay of messages in later channels
+        // mAX 2^192 == 2^6 * 2^18
```
ice09: don't understand the comment, can it be clarified?

contracts/contracts/RaidenMicroTransferChannels.sol
```diff
@@ -31,8 +31,11 @@ contract RaidenMicroTransferChannels {
...
+        uint192 deposit;
+
+        // UNIQUE for participants to prevent replay of messages in later channels
```
ice09: How is ensured that this is unique for participants?

contracts/contracts/RaidenMicroTransferChannels.sol
```diff
@@ -193,16 +198,12 @@ contract RaidenMicroTransferChannels {
...
     /// @param _sender The address that sends the tokens.
     /// @param _deposit The amount of tokens that the sender escrows.
     /// @param _data Receiver address in bytes.
-    function tokenFallback(
-        address _sender,
-        uint256 _deposit,
-        bytes _data)
-        external
-    {
+    function tokenFallback(address _sender, uint256 _deposit, bytes _data) external {
```
ice09: should be _sender_address?

contracts/contracts/RaidenMicroTransferChannels.sol
```diff
@@ -221,14 +222,11 @@ contract RaidenMicroTransferChannels {
...
+    /// @dev Creates a new channel between a sender and a receiver and transfers
+    /// the sender's token deposit to this contract, compatibility with ERC20 tokens.
     /// @param _receiver The address that receives tokens.
     /// @param _deposit The amount of tokens that the sender escrows.
-    function createChannelERC20(
-        address _receiver,
-        uint192 _deposit)
-        external
-    {
+    function createChannelERC20(address _receiver, uint192 _deposit) external {
```
ice09: also here: _receiver_address?

contracts/contracts/RaidenMicroTransferChannels.sol
```diff
@@ -252,12 +251,14 @@ contract RaidenMicroTransferChannels {
...
+    /// receiver needs a balance proof to immediately settle, sender triggers a challenge period.
     /// @param _receiver The address that receives tokens.
-    /// @param _open_block_number The block number at which a channel between the sender and receiver was created.
+    /// @param _open_block_number The block number at which a channel between the
+    /// sender and receiver was created.
     /// @param _balance The amount of tokens owed by the sender to the receiver.
     /// @param _balance_msg_sig The balance message signed by the sender.
-    function close(
+    function uncooperativeClose(
         address _receiver,
```
ice09: ..._address

contracts/contracts/RaidenMicroTransferChannels.sol
```diff
@@ -275,13 +276,15 @@ contract RaidenMicroTransferChannels {
...
     /// @param _receiver The address that receives tokens.
-    /// @param _open_block_number The block number at which a channel between the sender and receiver was created.
+    /// @param _open_block_number The block number at which a channel between the
+    /// sender and receiver was created.
     /// @param _balance The amount of tokens owed by the sender to the receiver.
     /// @param _balance_msg_sig The balance message signed by the sender.
     /// @param _closing_sig The hash of the signed balance message, signed by the receiver.
-    function close(
+    function cooperativeClose(
         address _receiver,
```
ice09: ..._address

contracts/contracts/RaidenMicroTransferChannels.sol
```diff
@@ -304,7 +307,8 @@ contract RaidenMicroTransferChannels {
     /// @dev Function for getting information about a channel.
     /// @param _sender The address that sends tokens.
     /// @param _receiver The address that receives tokens.
-    /// @param _open_block_number The block number at which a channel between the sender and receiver was created.
+    /// @param _open_block_number The block number at which a channel between the
+    /// sender and receiver was created.
     /// @return Channel information (unique_identifier, deposit, settle_block_number, closing_balance).
     function getChannelInfo(
         address _sender,
```
ice09: ..._address

contracts/contracts/RaidenMicroTransferChannels.sol
```diff
@@ -317,39 +321,42 @@ contract RaidenMicroTransferChannels {
...
-        uint32 _open_block_number)
-        external
-    {
+    /// @param _open_block_number The block number at which a channel between
+    /// the sender and receiver was created.
+    function settle(address _receiver, uint32 _open_block_number) external {
         bytes32 key = getKey(msg.sender, _receiver, _open_block_number);
 
+        // Make sure an uncooperativeClose has been initiated
         require(closing_requests[key].settle_block_number != 0);
```
ice09: didn't we use >0 always instead of !=0?

contracts/contracts/RaidenMicroTransferChannels.sol
```diff
@@ -317,39 +321,42 @@ contract RaidenMicroTransferChannels {
...
+    /// @param _sender The address that sends tokens.
     /// @param _receiver The address that receives tokens.
     /// @param _deposit The amount of tokens that the sender escrows.
-    function createChannelPrivate(
-        address _sender,
-        address _receiver,
-        uint192 _deposit)
-        private
-    {
+    function createChannelPrivate(address _sender, address _receiver, uint192 _deposit) private {
```
ice09: ..._address

contracts/contracts/RaidenMicroTransferChannels.sol
```diff
@@ -623,11 +569,7 @@ contract RaidenMicroTransferChannels {
         return string(bytesStringTrimmed);
     }
 
-    function addressToString(address x)
-        internal
-        pure
-        returns (string)
-    {
+    function addressToString(address x) internal pure returns (string) {
```
ice09: wouldn't call it "x"

contracts/tests/test_raidenchannels.py
```diff
@@ -285,15 +285,15 @@ def test_close_call(web3, chain, contract, channel):
...
-        contract.transact({'from': A}).close(receiver, open_block_number, balance, balance_msg_sig)
+        contract.transact({'from': A}).uncooperativeClose(receiver, open_block_number, balance, balance_msg_sig)
     with pytest.raises(tester.TransactionFailed):
-        contract.transact({'from': sender}).close(A, open_block_number, balance, balance_msg_sig)
+        contract.transact({'from': sender}).uncooperativeClose(A, open_block_number, balance, balance_msg_sig)
 
     # Cannot close if arguments not correct
     with pytest.raises(ValueError):
-        contract.transact({'from': sender}).close(receiver, open_block_number, balance)
+        contract.transact({'from': sender}).initChallengePeriod(receiver, open_block_number, balance)
```
ice09: since this is different from the code before, please double check, should it be initChallenge first and settleChannel then, is this correct for the test logic (might be, just not sure).

loredanacirstea: Tests will be re-written and re-checked, hopefully better documented soon. The previous test with `close` was an attempt to make sure the number of arguments was correct. That test doesn't make sense anymore anyway because we have overloading. Current test just makes sure `initChallengePeriod` is private.
Will leave this like it is for now and I'll have a test dedicated PR.

contracts/contracts/RaidenMicroTransferChannels.sol
loredanacirstea: Replacing this with
```
// Used in creating a unique identifier for the channel between a sender and receiver.
// Supports creation of multiple channels between the 2 parties and prevents
 // replay of messages in later channels
```

contracts/contracts/RaidenMicroTransferChannels.sol
loredanacirstea: Yes, didn't notice that the comment is very weird. The idea was to use the smallest `uint` size possible for `deposit`. 
10^8 * 10^18 max tokens (supply) -> 24 bytes -> uint192

Changing comment to:
```
// uint192 is the maximum uint size needed for deposit based on a 10^8 * 10^18 token totalSupply
```

PR #183, opened 2017-11-27
Contract - transfer tokens after state change

loredanacirstea:
Ensure Checks-Effects-Interactions Pattern to avoid race conditions

Channel data has to be deleted in `settleChannel` before the token transfers.
I don't think there can be a race condition in `topUpPrivate`, but at least we fail faster topUpPrivate has a lot of checks.


LefterisJP:
@loredanacirstea Looks good to me. I am merging.

PR #184, opened 2017-11-27
Contract - remove deposit from ChannelToppedUp & save gas

loredanacirstea:
We do not need `_ deposit` in `ChannelToppedUp` event. We already have `_added_deposit`.
This saves gas.


LefterisJP:
Looks good. Saving a tiny bit of gas.

PR #186, opened 2017-11-27
Increase challenge_period type size from uint8 to uint32

loredanacirstea:
`uint8` -> `max 2**8 * ~15sec / 60 = 64 min`
`uint16` -> `max 2**16 * 15 / 60 = 16384 min = 273h = 11.37 days`
`uint32` -> `max 2**16 * 15 / 60 = 45654 days`

uint8 was too small. It was chosen for decreasing gas - which is actually debatable and should be tested (EVM works with 256). Even so, you may want larger values without modifying the code.

Challenge period is anyway set at deployment, so you can make it as small as you like.

Also, it's used like this: (note block number type)
```
closing_requests[key].settle_block_number = uint32(block.number) + challenge_period;
```

### DIFF
--- a/contracts/contracts/RaidenMicroTransferChannels.sol
+++ b/contracts/contracts/RaidenMicroTransferChannels.sol
@@ -12,7 +12,7 @@ contract RaidenMicroTransferChannels {
 
     address public owner_address;
     address public token_address;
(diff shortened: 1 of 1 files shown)


LefterisJP:
Looks good to me.

PR #197, opened 2017-11-27
Explicit visibility and code style fixes

loredanacirstea:
Channel data is already visible with `getChannelInfo` anyway.

Mentioned in https://github.com/raiden-network/microraiden/issues/134#issuecomment-347144899

### DIFF
--- a/contracts/contracts/RaidenMicroTransferChannels.sol
+++ b/contracts/contracts/RaidenMicroTransferChannels.sol
@@ -24,10 +24,10 @@ contract RaidenMicroTransferChannels {
 
     string constant prefix = "\x19Ethereum Signed Message:\n";
 
-    Token token;
+    Token public token;
 
-    mapping (bytes32 => Channel) channels;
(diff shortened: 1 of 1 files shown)


contracts/contracts/RaidenMicroTransferChannels.sol
```diff
@@ -42,8 +42,8 @@ contract RaidenMicroTransferChannels {
     }
 
     struct ClosingRequest {
-        uint32 settle_block_number;
         uint192 closing_balance;
+        uint32 settle_block_number;
```
LefterisJP: @loredanacirstea Curious: Why change the order of the struct members here? Alphabetical?

loredanacirstea: Just for consistency with the above (sizes). Also mentioned as a code style improvement in https://github.com/raiden-network/microraiden/issues/134#issuecomment-347144899 and I said why not ..

Issue #193, opened 2017-11-27
Fix how closing_sig is handled after latest changes


loredanacirstea:
Possible reply attack problem:

From https://github.com/raiden-network/microraiden/issues/134#issuecomment-347144899
"
Biggest issue: verifyBalanceProof

The balanceProof only consist of: _receiver_address, _open_block_number, _balance. There is no information included on what you are signing, only data. The balanceProof is in this contract already used for 2 things. The sender verifies the new balance and the receiver use this to sign it wants to cooperatively close the channel. This is bad and opens a door for potential replay attacks:

- If any other contract just uses this 3 values for something else (unlikely)
- If there are two channels between address1 and address2 in both directions opened on the same block, they can be used vise versa potential with wrong meaning, if this would be implemented where receiver on the cooperative close is the sender.
- !!At the moment, the close request for cooperative close is just a balanceProof signed by the receiver and the receiving address is the receiver, this message is so not tied to any channel (sender, receiver) because it does not include a sender and can be used to close any channel with the given receiver!!
fix: Include some kind of message id that encodes the meaning, either cooperativeCloseChannel or balanceProof
- there is another contract deployed which has a channel of the same participants at the same block maybe for example for a different token, the signed messages can be used there.
fix: include the contract address in the signed message
"

loredanacirstea:
I do see the (receiver, block, balance) combination reproducible in different (very possible) circumstances:
- another sender (might be the same person, with a differet account)
- another contract 
  - as you said, same receiver/service provider uses another contract with a different token
  - another contract version, receiver wants senders to use new version while slowly deprecating the older one (waiting for senders to close channels)
- another network (receiver also deploys on a testnet for testing)


**Documenting changes:**

The initial logic was `closing_sig` being `balance_msg_sig` (sender-signed balance proof) signed by the receiver. This was changed on purpose at some point.

1. Initial logic:

https://github.com/raiden-network/microraiden/blob/8d6aac49a874138ef25d6aee4276a49799d1532f/contracts/contracts/RaidenMicroTransferChannels.sol#L82
```
# balanceMessageHash
return sha3(_receiver, _open_block_number, _balance, address(this));

# closingAgreementMessageHash
return sha3(_balance_msg_sig);

# separate handling for _balance_msg_sig and closing_sig
```


2. issue that triggered the change: https://github.com/raiden-network/microraiden/issues/52 , where we actually discussed the issue.
[...]

PR #198, opened 2017-11-27
Fix signature logic - prevent replay attack issue

loredanacirstea:
Fixes https://github.com/raiden-network/microraiden/issues/193
Discovered in https://github.com/raiden-network/microraiden/issues/134#issuecomment-347144899

- Include contract address in `_balance_msg_sig`
- Include contract & sender info in `_closing_sig` by just signing `keccak256(_balance_msg_sig)`
- Remove redundant signature length checks. This is done in `ECVerify.ecverify`.

### DIFF
--- a/contracts/tests/test_channel_close.py
+++ b/contracts/tests/test_channel_close.py
@@ -2,6 +2,7 @@ import pytest
 from ethereum import tester
 from utils import sign
 from tests.utils import balance_proof_hash
+from utils.utils import sol_sha3
 from tests.fixtures import (
     contract_params,
     owner_index,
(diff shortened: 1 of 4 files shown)


contracts/tests/test_channel_close.py
```diff
@@ -40,7 +41,12 @@ def test_close_call(get_accounts, uraiden_instance, token_instance, get_channel)
     (sender, receiver, open_block_number) = channel
     balance = channel_deposit - 10
 
-    balance_message_hash = balance_proof_hash(receiver, open_block_number, balance)
+    balance_message_hash = balance_proof_hash(
+        receiver,
+        open_block_number,
+        balance,
+        uraiden_instance.address
```
LefterisJP: general naming question. I see that `uraiden_instance` as seen [here](https://github.com/raiden-network/microraiden/blob/master/contracts/tests/test_uraiden.py#L69) and in the [fixture](https://github.com/raiden-network/microraiden/blob/master/contracts/tests/fixtures_uraiden.py#L54) is actually the `uraiden_contract` and not an instance of the uraiden_app as the name may imply. It's used quite a lot in tests so this should not be an urgent refactor but I would call it `uraiden_contract` and not just instance as it can be confusing.

loredanacirstea: I use
`_instance` as an off-the-shelf deployed contract.
`_contract` as a function that can take other parameters (token, txn data)
`get_*_contract` for a lower level setup

Agree that the name is not great. Yes, will leave it like this for now, but noted.

Issue #181, opened 2017-11-27
eth_signTypedData - add verifyBalanceProofPrepend() ; prepend schema hash


loredanacirstea:
We need to add a second method for calculating the message hash `verifyBalanceProofPrepend` 
The second method is not yet implemented ethereum/EIPs#712 (comment) 
Example code: https://github.com/0xProject/EIPs/blob/01dfc0f9a4122d8ad8817c503447cab8efa8a6c4/EIPS/eip-signTypedData.md#pseudocode-examples

Because EIP712 is not standardized yet, a solution is to put the signature-related logic in a contract-library. The contract-library address is modifiable by the uRaiden contract owner.

PR #200, opened 2017-11-28
Extract balance message hash logic for EIP712 in a separate contract

loredanacirstea:
merge after https://github.com/raiden-network/microraiden/pull/198

fixes https://github.com/raiden-network/microraiden/issues/181 - you can change the balance proof message hash logic at any point.

EIP712 is not standardized and can have breaking changes.

- adds a setter in the uRaiden contract for setting the `URaidenEIP712HelperContract` address, callable only by the `RaidenMicroTransferChannels` contract owner
- exposes `getMessageHash` as a `pure` function, used in the uRaiden `verifyBalanceProof` function

### DIFF
--- a/contracts/contracts/URaidenEIP712HelperContract.sol
+++ b/contracts/contracts/URaidenEIP712HelperContract.sol
@@ -1,0 +1,34 @@
+pragma solidity ^0.4.17;
+
+contract URaidenEIP712HelperContract {
+    /// @dev Returns the message hash used for creating the balance proof.
+    /// Used in the RaidenMicroTransferChannels contract.
+    /// Should be kept up-to-date with eth_signTypedData https://github.com/ethereum/EIPs/pull/712.
+    /// @param _receiver_address The address that receives tokens.
+    /// @param _open_block_number The block number at which a channel between the
+    /// sender and receiver was created.
+    /// @param _balance The amount of tokens owed by the sender to the receiver.
+    /// @param _contract RaidenMicroTransferChannels contract address
+    /// @return Hash of the message composed from the above parameters.
+    function getMessageHash(
+        address _receiver_address,
+        uint32 _open_block_number,
+        uint192 _balance,
+        address _contract)
+        public
+        pure
+        returns (bytes32)
+    {
+        // Used in the RaidenMicroTransferChannels contract
+        // The variable names from below will be shown to the sender when signing
+        // the balance proof, so they have to be kept in sync with the Dapp client.
+        // The hashed strings should be kept in sync with this function's parameters
+        // (variable names and types)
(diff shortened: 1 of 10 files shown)


contracts/contracts/URaidenEIP712HelperContract.sol
```diff
@@ -0,0 +1,34 @@
+pragma solidity ^0.4.17;
+
+contract URaidenEIP712HelperContract {
```
LefterisJP: ###Naming Suggestion###

We call the microraiden contract `RaidenMicroTransferChannels.sol` so the library contract should also use `micro` and not `u`. So I would suggest something like `MicoRaidenEIP712Helper`. or `MicroRaidenEIP712Signer`. No need to keep the `contract` part in the name. We know it's a contract.

contracts/contracts/URaidenEIP712HelperContract.sol
```diff
@@ -0,0 +1,34 @@
...
+    function getMessageHash(
+        address _receiver_address,
+        uint32 _open_block_number,
+        uint192 _balance,
+        address _contract)
+        public
+        pure
+        returns (bytes32)
+    {
+        // Used in the RaidenMicroTransferChannels contract
```
LefterisJP: Remove comment. Already written above.

contracts/contracts/URaidenEIP712HelperContract.sol
```diff
@@ -0,0 +1,34 @@
...
+        // The variable names from below will be shown to the sender when signing
+        // the balance proof, so they have to be kept in sync with the Dapp client.
+        // The hashed strings should be kept in sync with this function's parameters
+        // (variable names and types)
+        bytes32 message_hash = keccak256(
+          keccak256('address receiver', 'uint32 block_created', 'uint192 balance', 'address contract'),
```
LefterisJP: The test [here](https://github.com/raiden-network/microraiden/pull/200/files#diff-00b1cc92080ad0cb14ac960cb278f551R155) guarantees that they are kept in sync, right? If not then we need such a test.

contracts/contracts/URaidenEIP712HelperContract.sol
loredanacirstea: `MicoRaidenEIP712Helper` it is then. I tried a couple of names, but I knew you are better at this and will probably find a good one hahah.

contracts/contracts/URaidenEIP712HelperContract.sol
loredanacirstea: Tests have been added here https://github.com/raiden-network/microraiden/pull/200/files#diff-bd130ba0c786d648c876d74356582f11R50

PR #201, opened 2017-11-28
Closing signature and balance proof replay protection (Python-side)

devium:
Synced with contract changes introduced in #198


loredanacirstea:
Looks good.
Implements the last contract-related changes: creating the balance proof, how we do the closing signature now, latest topUp event args.

Issue #135, opened 2017-11-09
Documenting the external audit of the smart contract


loredanacirstea:
Scope:
- RaidenMicroTransferChannels contract: https://github.com/raiden-network/microraiden/blob/master/contracts/contracts/RaidenMicroTransferChannels.sol
- MicroRaidenEIP712Helper contract: https://github.com/raiden-network/microraiden/blob/master/contracts/contracts/MicroRaidenEIP712Helper.sol
- ECVerify library: https://github.com/raiden-network/microraiden/blob/master/contracts/contracts/lib/ECVerify.sol
- Token.sol: https://github.com/raiden-network/microraiden/blob/master/contracts/contracts/Token.sol

Outcome:
1. redundant: either `token` or  `token_address`, `isToken`, `prefix`
fixed in https://github.com/raiden-network/microraiden/pull/196

Reviewed version for issues below: https://github.com/raiden-network/microraiden/tree/81920e4162da8b00ba54212de619668d42c11120

2. tl;dr In this contract, the sender needs to fully trust the owner.
Owner can be malicious / his account can be compromised and change `MicroRaidenEIP712Helper.getMessageHash` by deploying another contract with malicious code  and changing the address in `RaidenMicroTransferChannels` with `setEip712HelperContract`.

Possible scenarios:
- `getMessageHash` is changed -> channels are locked, because the sender's signature from `_balance_msg_sig` cannot be recovered.
- owner colludes with 1 or > receivers to modify the function so that they can get more tokens out of the channels than they are owed (amount of tokens limited by the actual channel deposit)

The receiver can use any hash of a signed message from the sender, to be returned in a hardcoded way in the EIP712 helper.

(e.g. :  in `RaidenMicroTransferChannels.cooperativeClose` -> `verifyBalanceProof`,  `getMessageHash` can return the hash corresponding to `_balance - 100` instead of `_balance`. This means the receiver can provide an authentic `_balance_msg_sig` with `balance = _balance - 100`, but use `_balance` as an argument for `cooperativeClose`.)

Possible solution:

Adding a function to change the owner would be important. Specially if the owner address gets compromised.
I would put another role. `eip712updater` so you can set it to `0x0` when the eip712 standard is stabilized. And still maintain the option to `setLatestVersionAddress`. Needed methods: `changeOwner`, `changeEip712updater` .. In the constructors both vars can be set to msg.sender

Having owner and eip712updater diferently allows to cancel the possibility to change the eip712 helper and still update the version.

[...]

PR #209, opened 2017-11-29
Decentralise: remove owner, eip712 helper, latest address setter

loredanacirstea:
In light of the external audit findings documented in https://github.com/raiden-network/microraiden/issues/135, issue 2:
> In this contract, the sender needs to fully trust the owner.
Owner can be malicious / his account can be compromised and change MicroRaidenEIP712Helper.getMessageHash by deploying another contract with malicious code and changing the address in RaidenMicroTransferChannels with setEip712HelperContract.

we decided to remove the owner and functions that the owner used to change the contract state.

- removed `owner_address`
- removed `setLatestVersionAddress`
- removed `setEip712HelperContract`
- removed `MicroRaidenEIP712Helper` and included the `getMessageHash` logic into the main contract again (we can re-deploy after EIP712 changes)


loredanacirstea:
@LefterisJP ,

My concern about any breaking change from EIP712 at this point is: it is not standardized, we cannot be sure that the current (experimental) implementation will be degraded gracefully. I implemented the upgradable message hash exactly due to this reasoning.

andrevmatos:
@loredanacirstea one point about it is that we're not tied to the EIP712 implementation/revision. We could make the client/server identify when contract uses eip's rev.1 or 2, and use old hashing algorithm. Only drawback is that in these legacy cases, Metamask will present user with an opaque hash for signing (same as with Parity or any client which doesn't support either EIP712 proposals), and user will need to check it old style (maybe a UI link on how to generate that same hash externally and verify it?). 
- Server side, nothing changes (besides using the correct algorithm for the currently configured contract). 
- Client side, we'll support in the official contracts most recent implementation, so web UI will always show as much info as possible. m2m also doesn't care, as no data is "presented" for signing, and client code implements the hashing algorithm directly.

loredanacirstea:
@andrevmatos , agreed. This approach is better than having an upgradable contract with possible trust and centralization issues. The idea with providing tools or explanation for how to verify old style, human-unreadable messages is indeed acceptable from my point of view, especially that this is the current way of doing things.

LefterisJP:
> one point about it is that we're not tied to the EIP712 implementation/revision. We could make the client/server identify when contract uses eip's rev.1 or 2 (maybe a small constant?)

The version of the contract should be sufficient for this, right?

 
> and user will need to check it old style (maybe a UI link on how to generate that same hash externally and verify it?).

@andrevmatos Can you elaborate a bit on what old-style is? How would that work?

andrevmatos:
@LefterisJP old-style is user be oriented to externally check that the data it's supposed to agree really produces the hash he's about to be asked to sign. EIP712 is just a way for the wallet (Metamask) to offer itself (in a standardized way) as a trusted 3rd party to verify, show, hash and sign these data. In legacy code, as the wallet won't (possibly) be able to do it for the user (or hasn't support for the eip yet), the user will be shown the data, and linked to simple instructions on how to verify it. Maybe an on-chain contract with only const functions for the known hashing formats, and etherscan forms to interact with it. Of course we don't expect the user to do this everytime, users may do that only if in doubt the data to be signed is correct in a trustless manner (or at least trusting things out of receiver's control, as blockchain).

PR #210, opened 2017-11-30
Add temporary bug bounty release channel deposit limit

loredanacirstea:
Limit token deposits for the bug bounty Mainnet release.

Thought about taking token.decimals() into consideration, but decided to go with the simple solution. This function is not a requirement for erc20. Additionally, the scope is to preventively limit loss - hard to calculate for other tokens.

### DIFF
--- a/contracts/contracts/RaidenMicroTransferChannels.sol
+++ b/contracts/contracts/RaidenMicroTransferChannels.sol
@@ -15,6 +15,10 @@ contract RaidenMicroTransferChannels {
     // Contract semantic version
     string public constant version = '0.1.0';
 
+    // We temporarily limit total token deposits in a channel to 100 tokens with 18 decimals.
+    // This is just for the bug bounty release.
+    uint256 public channel_deposit_bugbounty_limit = 10 ** 18 * 100;
+
     Token public token;
 
     mapping (bytes32 => Channel) public channels;
@@ -310,6 +314,8 @@ contract RaidenMicroTransferChannels {
     /// @param _receiver_address The address that receives tokens.
     /// @param _deposit The amount of tokens that the sender escrows.
     function createChannelPrivate(address _sender_address, address _receiver_address, uint192 _deposit) private {
+        require(_deposit <= channel_deposit_bugbounty_limit);
+
         uint32 open_block_number = uint32(block.number);
 
         // Create unique identifier from sender, receiver and current block number
@@ -344,6 +350,7 @@ contract RaidenMicroTransferChannels {
 
         require(channels[key].deposit > 0);
         require(closing_requests[key].settle_block_number == 0);
+        require(channels[key].deposit + _added_deposit <= channel_deposit_bugbounty_limit);
 
         channels[key].deposit += _added_deposit;
         assert(channels[key].deposit > _added_deposit);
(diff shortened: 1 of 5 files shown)


contracts/contracts/RaidenMicroTransferChannels.sol
```diff
@@ -15,6 +15,10 @@ contract RaidenMicroTransferChannels {
     // Contract semantic version
     string public constant version = '1.0.0';
 
+    // We temporarily limit total token deposits in a channel to 100 tokens with 18 decimals.
+    // This is just for the bug bounty release.
```
LefterisJP: Please also mention here in a comment that this is just for RDN and with its current (as of 30/11/2017) price and should not be considered to be the same for other tokens.

contracts/contracts/RaidenMicroTransferChannels.sol
```diff
@@ -15,6 +15,10 @@ contract RaidenMicroTransferChannels {
...
+    uint256 public channel_deposit_bugbounty_limit = 10 ** 18 * 100;
```
LefterisJP: Please make it a constant since it's not supposed to ever be changed for the lifetime of this contract

LefterisJP:
@loredanacirstea I think you also need to adjust the tests for this and also write a test that checks that we fail to create a channel for more than 100 tokens of value.

contracts/contracts/RaidenMicroTransferChannels.sol
```diff
@@ -310,6 +316,8 @@ contract RaidenMicroTransferChannels {
     /// @param _receiver_address The address that receives tokens.
     /// @param _deposit The amount of tokens that the sender escrows.
     function createChannelPrivate(address _sender_address, address _receiver_address, uint192 _deposit) private {
+        require(_deposit <= channel_deposit_bugbounty_limit);
```
czepluch: Didn't look at it thoroughly, but doesn't this still allow you to top up a channel with as many tokens as you want?

czepluch: Ah, nevermind. The 'updateInternalBalanceStructs' take care of this.

loredanacirstea: there is another check at topup

contracts/contracts/RaidenMicroTransferChannels.sol
```diff
@@ -344,6 +352,7 @@ contract RaidenMicroTransferChannels {
 
         require(channels[key].deposit > 0);
         require(closing_requests[key].settle_block_number == 0);
+        require(channels[key].deposit + _added_deposit <= channel_deposit_bugbounty_limit);
```
ice09: that's overflow, no? If I test this, I can submit `_added_deposit=2**192-1` and with a deposit of just 1 it doesn't fail.
```
    function addall() {
        uint192 add1 = 10;
        uint192 add2 = 2**192-1;
        require(add1 + add2 <= 100);
    }
```

loredanacirstea: If it is an overflow, it will fail at `assert(channels[key].deposit > _added_deposit);` , right?
So you tested and it did not fail?

ice09: yes, that's right, it would fail then.

PR #213, opened 2017-11-30
Contract - set correct semantic version.

loredanacirstea:


LefterisJP:
looks good now. Will merge.

PR #218, opened 2017-11-30
Fix for wrong token address call in Python lib

devium:
After the recent renaming of the token address function from `token_address()` to `token()`, Python code using that function wasn't updated.

I included some refactoring so at least on the client side, this is now also covered by the tests.


microraiden/microraiden/client/client.py
```diff
@@ -73,34 +67,32 @@ def __init__(
         if not web3:
             if channel_manager_proxy:
                 self.web3 = channel_manager_proxy.web3
+                self.channel_manager_address = channel_manager_proxy.address
             elif token_proxy:
                 self.web3 = token_proxy.web3
             else:
-                if not rpc:
```
pcppcp: Won't this be problematic? What if you want to use ethereum node that's not on the localhost (or Infura or something)?

devium: Then pass a `Web3(RPCProvider('somewhere.else', 8545))` as `web3`.

This doesn't really add any complexity but simplifies the ctor signature quite a bit.

loredanacirstea: @pcppcp , can this be closed?

pcppcp: ah, I see. I am also in favor of having ctor use less params.

PR #219, opened 2017-11-30
Contract - require challenge_period >= 500

loredanacirstea:
As indicated in the external audit, point 7 https://github.com/raiden-network/microraiden/issues/135

### DIFF
--- a/contracts/contracts/RaidenMicroTransferChannels.sol
+++ b/contracts/contracts/RaidenMicroTransferChannels.sol
@@ -74,12 +74,13 @@ contract RaidenMicroTransferChannels {
 
     /// @dev Constructor for creating the uRaiden microtransfer channels contract.
     /// @param _token_address The address of the Token used by the uRaiden contract.
-    /// @param _challenge_period A fixed number of blocks representing the challenge period
+    /// @param _challenge_period A fixed number of blocks representing the challenge period.
+    /// We enforce a minimum of 500 blocks waiting period.
     /// after a sender requests the closing of the channel without the receiver's signature.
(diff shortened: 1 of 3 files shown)


contracts/contracts/RaidenMicroTransferChannels.sol
```diff
@@ -74,12 +74,13 @@ contract RaidenMicroTransferChannels {
...
     /// @param _token_address The address of the Token used by the uRaiden contract.
-    /// @param _challenge_period A fixed number of blocks representing the challenge period
+    /// @param _challenge_period A fixed number of blocks representing the challenge period.
+    /// We enforce a minimum of 500 blocks waiting period.
     /// after a sender requests the closing of the channel without the receiver's signature.
     function RaidenMicroTransferChannels(address _token_address, uint32 _challenge_period) public {
         require(_token_address != 0x0);
         require(addressHasCode(_token_address));
-        require(_challenge_period > 0);
+        require(_challenge_period >= 500);
```
LefterisJP: Since this:

1. is gonna be deployed by us
2. the challenge period is not editable after deployment
3. Any further upgrades will require more thinking around the challenge period but also a new contract deployment.

Why not remove this from the constructor, make it a constant in the contract and remove this require? Will also save us some gas wherever we read this from.

loredanacirstea: > is gonna be deployed by us
We should support people wanting to deploy on testnet. I think this should be kept like this, so testing is easier + we would need more changes to the tests at this point.

LefterisJP: fair enough

PR #220, opened 2017-11-30
Contract - check constructor arg is a Token address

loredanacirstea:
Recommended in point 8 of the external audit: https://github.com/raiden-network/microraiden/issues/135

### DIFF
--- a/contracts/tests/test_uraiden.py
+++ b/contracts/tests/test_uraiden.py
@@ -62,7 +62,7 @@ def test_uraiden_init(
     with pytest.raises(tester.TransactionFailed):
         get_uraiden_contract([token.address, challenge_period_min - 1])
     with pytest.raises(tester.TransactionFailed):
-        get_uraiden_contract([token.fake_token, challenge_period_min])
+        get_uraiden_contract([fake_token.address, challenge_period_min])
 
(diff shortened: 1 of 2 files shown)


contracts/tests/test_uraiden.py
```diff
@@ -56,6 +57,8 @@ def test_uraiden_init(
         get_uraiden_contract([A, 100])
     with pytest.raises(tester.TransactionFailed):
         get_uraiden_contract([token.address, 0])
+    with pytest.raises(tester.TransactionFailed):
+        get_uraiden_contract([fake_token.address, 100])
```
czepluch: Maybe make the `challenge_period` valid here to emphasize that it's the address that fails and not the challenge period. And make sure to do that in the other calls too, where it's not the challenge period that's supposed to fail but the address.

loredanacirstea: It is valid. Will change it after rebasing the other PR

czepluch: I know it's valid now, but since I saw the other PR too, I just wanted to make sure that you're aware of this. Will approve for now then.

LefterisJP:
looks fine to me

PR #239, opened 2017-12-01
Fix topup event argument names

konradkonrad:
The argument name for a top up event `_deposit` changed to `_added_deposit`.

### DIFF
--- a/microraiden/microraiden/client/client.py
+++ b/microraiden/microraiden/client/client.py
@@ -160,7 +160,7 @@ class Client:
 
         for e in topup:
             c = get_channel(e)
-            c.deposit = e['args']['_deposit']
+            c.deposit += e['args']['_added_deposit']
 
         for e in close:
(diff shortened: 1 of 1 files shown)


loredanacirstea:
@konradkonrad , @LefterisJP , we had `_added_deposit` & `_deposit` before. 
It should be:
`c.deposit += e['args']['_added_deposit']` now.

konradkonrad:
> It should be:
`c.deposit += e['args']['_added_deposit']` now.

addressed this and rebased

devium:
Just note: this might cause issues
1. when the deposit events are not replayed correctly before replaying the topup events or
2. when the topup events are replayed multiple times (node error, reorg, etc).

In the second case, clients might assume a higher capacity than they actually have, becoming stuck in a request loop with the server, since the client assumes that server just hasn't heard of the topup yet.

I don't think this is relevant right now, but might be something to keep in mind.

Issue #244, opened 2017-12-04
DefaultHTTPClient doesn't retry after invalid amount sent


devium:
Essentially, change this to `True` and cover it with a test:
https://github.com/raiden-network/microraiden/blob/57207be3f4c10cc4cbb44857b9d66d351259676c/microraiden/microraiden/client/default_http_client.py#L56

Problem:
When the client sends an invalid balance to the sever, for example because the client lost its most recent balance proof, it adopts the corrected balance proof from the server but does not repeat the request with the new balance.

devium:
The above PR hotfixes the bug but this should still be covered by a test.

PR #245, opened 2017-12-04
Allow retry on invalid amount

konradkonrad:
If the client is out of sync with the server and receives payment headers with `Rdn-Invalid-Amount` set, it should retry the request with an updated balance.


devium:
Looks fine. I'll leave #244 open though, for the test. We should definitely have one that goes through this kind of client/server exchange.

PR #250, opened 2017-12-05
added `on_http_response()` method to the http client

### LINKED ISSUE #249: Make server responses and status codes available in the client
Several possible ways to do this:
* Return the last (arguably most important) status code, header, and response body from the `run()` method
* Return a list of all status codes, headers, and response bodies involved in a single `run()` call from the `run()` method
* Introduce an `on_response()` callback that also returns a `bool` indicating a retry.

I slightly prefer the last option.

pcppcp:
fixes #249

### DIFF
--- a/microraiden/microraiden/client/http_client.py
+++ b/microraiden/microraiden/client/http_client.py
@@ -1,5 +1,5 @@
 import logging
-from typing import Any
+from typing import Tuple, Any
 
 import requests
 from eth_utils import encode_hex, decode_hex
@@ -61,7 +61,7 @@ class HTTPClient(object):
         if self.channel:
             self.close_channel(self.channel)
 
-    def _request_resource(self, requested_resource: str) -> (Any, bool):
+    def _request_resource(self, requested_resource: str) -> Tuple[Any, bool]:
         """
         Performs a simple GET request to the HTTP server with headers representing the given
         channel state.
@@ -78,6 +78,9 @@ class HTTPClient(object):
         url = self.make_url(requested_resource)
         response = requests.get(url, headers=HTTPHeaders.serialize(headers))
         headers = HTTPHeaders.deserialize(response.headers)
+
+        if self.on_http_response(requested_resource, response) is False:
+            return None, False  # user requested abort
 
         if response.status_code == requests.codes.OK:
             retry = self.on_success(response.content, headers.get('cost'))
@@ -106,7 +109,7 @@ class HTTPClient(object):
                 )
             return None, retry
         else:
-            return None, False
+            return None, True
 
     def on_init(self, requested_resource):
         pass
@@ -135,3 +138,10 @@ class HTTPClient(object):
             channel_manager_address: str
     ):
         return True
+
+    def on_http_response(self,
+                         resource: str,
+                         reply):
+        """Called whenever server returns a reply.
+        Return False to abort current request."""
+        return reply.status_code in (200, 402)
--- a/microraiden/microraiden/test/utils/client.py
+++ b/microraiden/microraiden/test/utils/client.py
@@ -1,5 +1,7 @@
+import types
+
 from microraiden.client import Channel
-from microraiden import Client
+from microraiden import Client, DefaultHTTPClient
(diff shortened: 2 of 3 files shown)


microraiden/microraiden/client/http_client.py
```diff
@@ -135,3 +138,10 @@ def on_payment_requested(
             channel_manager_address: str
     ):
         return True
+
+    def on_http_response(self,
```
devium: Two notes:
* Since we're using type hints (albeit in a bit of a half-assed fashion), please include `reply: Response` with `from requests import Response` here.
* (sorry for nitpicking) using the Raiden style guide we should start multiline function calls on the next line, like so:
```python
def on_http_response(
    self,
    resource: str,
    reply: Response
):
```
or 
```python
def on_http_response(
    self,
    resource: str,
    reply: Response):
```

microraiden/microraiden/client/http_client.py
```diff
@@ -106,7 +109,7 @@ def _request_resource(self, requested_resource: str) -> (Any, bool):
                 )
             return None, retry
         else:
-            return None, False
+            return None, True
```
devium: Under the new default behavior this code isn't reached, right?

If someone decides to override the check in `on_http_response()`, then this code would be reachable though. Should we really continue on errors like 500?

Maybe we should add another `on_server_error()` callback here which defaults to `True`, just to give users a maximum amount of configurability with sensible defaults.

microraiden/microraiden/test/utils/client.py
```diff
@@ -26,3 +28,13 @@ def close_all_channels_cooperatively(
                 if c.state != Channel.State.closed and c.receiver == receiver_addr]
     for channel in channels:
         close_channel_cooperatively(channel, privkey_receiver, contract_address, balance)
+
+
+def patch_on_http_response(default_http_client: DefaultHTTPClient, abort_on=[]):
+    def patched(self, resource, response):
+        self.last_response = response
+        return (response.status_code not in abort_on)
+    default_http_client.on_http_response = types.MethodType(
```
devium: Just a note (no changes needed):

`pytest` has the `monkeypatch` fixture (of type `_pytest.monkeypatch.Monkeypatch`) which would be a nice alternative to these `types.MethodType(...)` calls. For example I use this to mock `click` prompt inputs:

```python
def mock_prompt(monkeypatch: MonkeyPatch, input_sequence: Iterator[str]):
    def patched_prompt(*args, **kwargs):
        value = next(input_sequence)
        print('{}: {}'.format(args[0], value))
        return value

    monkeypatch.setattr(click, 'prompt', patched_prompt)
    monkeypatch.setattr(click, 'confirm', patched_prompt)
```

It automatically cleans the monkeypatching up after the test and allows to patch built-in methods too.

pcppcp: I've just tried that and it doesn't seem to be working with instances.

```python
def patched(self, , response):
          self.last_response = resp
          return False
monkeypatch.setattr(default_http_client, 'on_http_response', patched)
```

fails with 
`TypeError: patched() missing 1 required positional argument: 'response'`

microraiden/microraiden/client/http_client.py
pcppcp: There'll always be weird use cases, for example a scenario where user will want to have a simple HTTP auth to the client (for one reason or another), then the sequence will be 401 -> 402 -> 200.
It's anyways user's responsibility to recover if he is redefining this method..

microraiden/microraiden/client/http_client.py
pcppcp: Is there a tool to check/enforce type hints? My `mypy` integration is not really usable yet, I should spend some time making it work.
ad multine calls, I always forget them. I prefer less lines in the code so it's hard to keep it on mind if there's no linter to yell at me :-) It'll get better over the time.

microraiden/microraiden/test/utils/client.py
devium: Ok nevermind that comment then. It only really works with built-in modules I guess. With custom classes it gets ugly because you need to specify the import path.

PR #251, opened 2017-12-06
Improved (isolated) client test coverage, fixtures, and minor protocol changes

devium:
Sorry, this PR escalated a bit.

Changes include:
* Lots of the behavior in the `DefaultHTTPClient` were moved up to `HTTPClient` as a reasonable default.
* Isolated (without a doggo proxy) client tests verifying protocol compliance, including #244.
* New pytest option fixtures:
    * `--no-clean-channels` sets `clean_channels` to `False`: Prevents channels from all channels between `sender_address` and `receiver_address` being closed before and after each test. Unless you're very patient, you should probably use this on any real testnet.
    * ~`--no-verbose` sets `verbose` to `False`: Suppresses debug output from the client. Currently not used in the `channel_manager` fixture but might be worth thinking about.~
* Refactoring of `DefaultHTTPClient` and `HTTPClient` to keep logic a lot simpler, more modular, and easily extendible. Most notably, `on_payment_requested` is only called when no other error is present in the server's response. In the case of `on_invalid_amount`, for example, the default implementation calls `on_payment_requested` explicitly.


microraiden/requirements.txt
```diff
@@ -9,6 +9,7 @@ ethereum-tester
 web3==3.11.1
 gevent
 click
+requests_mock
```
pcppcp: Maybe the time has come to have `requirements.txt` and `requirements-dev.txt`

microraiden/microraiden/test/test_http_client.py
```diff
@@ -1,47 +1,212 @@
...
+    request = server_mock.request_history[0]
+    assert request.path == '/something'
+    assert request.method == 'GET'
+    assert request.headers['RDN-Contract-Address'] == channel_manager_contract_address
+
+    # Second cycle, pay price but receive error.
+    request = server_mock.request_history[1]
+    assert request.path == '/something'
+    assert request.method == 'GET'
+    assert request.headers['RDN-Contract-Address'] == channel_manager_contract_address
```
pcppcp: Why not use const definitions from `header.py` here?

microraiden/microraiden/client/http_client.py
```diff
@@ -83,65 +84,84 @@ def _request_resource(self, requested_resource: str) -> Tuple[Any, bool]:
...
     def on_init(self, requested_resource):
-        pass
+        log.info('Starting request loop for resource at {}.'.format(requested_resource))
 
     def on_exit(self):
         pass
 
-    def on_success(self, resource, cost: int):
+    def on_success(self, resource, cost: int) -> bool:
+        log.info('Resource received.')
```
pcppcp: Infos like these should be changed to debug or removed completely.

http://www.catb.org/esr/writings/taoup/html/ch01s06.html#id2878450

microraiden/microraiden/test/test_http_client.py
```diff
@@ -1,47 +1,212 @@
...
+        receiver_address
 ):
+    # Simulate a lost balance signature.
+    client = default_http_client.client
+    channel = client.get_suitable_channel(receiver_address, 10, initial_deposit=lambda x: 2 * x)
+    channel.create_transfer(3)
+    lost_balance_sig = channel.balance_sig
+    channel.balance = 0
+
+    with requests_mock.mock() as server_mock:
```
pcppcp: Good idea but we still need to have client-proxy integration tests :)

microraiden/microraiden/test/fixtures/client.py
```diff
@@ -62,24 +64,26 @@ def client(
         sender_privkey: str,
         client_contract_proxy: ChannelContractProxy,
         client_token_proxy: ContractProxy,
-        datadir: str
+        datadir: str,
+        verbose: bool,
+        clean_channels: bool,
+        receiver_privkey: str
```
pcppcp: I understand why you need this here, but isn't there a better place to do a channel cleanup? i.e. when tearing down `ChannelManager`?

microraiden/microraiden/test/fixtures/web3.py
```diff
@@ -192,7 +212,15 @@ def channel_manager_proxy_factory(privkey):
...
             use_tester
         )
-    return channel_manager_proxy_factory
+    # Everything is deployed. Make snapshot.
+    if use_tester:
+        snapshot_id = web3.testing.snapshot()
+
+        yield channel_manager_proxy_factory
+
+        web3.testing.revert(snapshot_id)
```
pcppcp: This doesn't make sense. The chain will be reverted when all tests finish?

microraiden/microraiden/test/fixtures/client.py
devium: There are some client tests that run without a channel manager. Are there channel manager tests that open channels without this client fixture?

microraiden/microraiden/test/test_http_client.py
devium: Sooon. Although many are already in there.

microraiden/microraiden/test/fixtures/web3.py
devium: Damn, didn't see the `session` scope. I guess I will make a `web3_snapshot` fixture that both `channel_manager` and `client` use. That explains why the tests are slower.

I had the issue that the channel manager reverted the tester chain before the client could clean up its channels. So I'll look into fixing that again.

microraiden/microraiden/test/test_http_client.py
devium: My semi-logical reasoning here was that I'm testing for protocol compatibility.

If the `header.py` definition is wrong then this test would still pass if using the `Munch`-based header.

microraiden/requirements.txt
devium: Will do.

microraiden/microraiden/client/channel.py
```diff
@@ -105,7 +105,7 @@ def topup(self, deposit):
         )
         self.client.web3.eth.sendRawTransaction(tx)
 
-        log.info('Waiting for topup confirmation event...')
+        log.debug('Waiting for topup confirmation event...')
```
pcppcp: just a comment: it's a good practice to always include data you are trying to debug in the log message (here the message can be `Waiting for topup confirmation event... tx=0xdead...beef`)

Issue #255, opened 2017-12-07
Transition to a stateless and concurrent HTTPClient


devium:
The `HTTPClient` could be made stateless to allow for easy concurrent requests. Request parameters would be passed to each `on_...()` callback and channel references would be stored per endpoint url (which makes it not quite so stateless). Race conditions should not take more than one additional cycle if the server can handle it properly.

For example, this works fine:
1. Client requests resource 1, with channel balance 0
1. Client also requests resource 2, with channel balance 0
1. Server responds with price 3 for resource 1
1. Client requests resource 1 with channel balance 3
1. Server responds with price 4 for resource 2
1. Client requests resource 2 with channel balance 7
1. Server accepts payment for resource 1, returns it
1. Server accepts payment for resource 2, returns it

This causes some additional cycles:
1. Client requests resource 1, with channel balance 0
1. Client also requests resource 2, with channel balance 0
1. Server responds with price 3 for resource 1
1. Client requests resource 1 with channel balance 3
1. Server responds with price 4 for resource 2
1. Client requests resource 2 with channel balance 7
1. Server receives payment for resource 2, signals `invalid_amount`, sends `last_balance` of 0
1. Server accepts payment for resource 1, returns it
1. Client requests resource 2 again, based on `last_balance==0`, sends channel balance 4
1. Server again receives the payment and signals `invalid_amount`, based on `last_balance` of 3
1. Client is getting pissed and requests again resource 2 with channel balance 7
1. Server accepts payment for resource 2, returns it

In general, concurrent requests will work, but keeping them synchronous speeds things up a little. It might also be possible to use more than one channel for concurrent requests.

In the long-term we should probably introduce sequence numbers for these requests. See also #195

devium:
On this note, it also makes sense to unify all `on_...()` callbacks to use the same signature:
`self, method: str, url: str, response: Response, **kwargs`
except for `on_init()` which does not have a response yet.
`method` and `**kwargs` would be included as part of @konradkonrad's suggestion in issue #187

RDN Header fields can easily be read, e.g. using `response.headers[HTTPHeaders.PRICE]`.

Issue #187, opened 2017-11-27
Support other HTTP methods than GET


devium:
Currently, µRaiden only supports `GET` requests to be performed by the client and forwarded by the server. When using REST APIs, however, you usually need more than that.

While other methods could be enclosed in the `GET` request body, the Python client currently has no easy way to include that body in the request, other than overriding the `_request_resource()` method.

It would be nice if either
* (easy, short term) The client would support arbitrary request bodies or
* (more complex, long term) Both client and server learn to natively handle and forward POST/PUT/PATCH/etc.

konradkonrad:
I think it would be really awesome, if
- the client could just expose a patched `requests` module (and support `microraiden_requests.get`, `microraiden_requests.post`, ... ). Ideally, that would allow developers to swap out `requests` from an off-the-shelf library and immediately speak microraiden...
- the server could just wrap any `flask` (or WSGI?) app.

For starters, a patched `requests.request(method, url, **kwargs)` could get us pretty far for the client.

pcppcp:
Fixed in #272 #258

PR #258, opened 2017-12-07
Client-side support for all HTTP methods

devium:
Again, a meatier PR but the individual commits are more narrowly scoped.

Changes:
* `DefaultHTTPClient` now supports all HTTP methods supported by `requests.request`, including a simple forwarding of `kwargs` (e.g. for request bodies).
* Callback method signatures are unified, with all of them taking the initial request parameters and request response, as described in #255 
* Introduced request-(or rather endpoint-)specific channel states in a `HTTPClient`, meaning concurrent requests should work in theory, but we cannot properly test this until the proxy implements concurrent requests. A single `HTTPClient` still uses a single `Client` object for all of its transfers, meaning it can only support transfers from a single address.
* A new `microraiden.requests` module that -- after an initial call to `microraiden.requests.init()` -- can be used just like the `requests` module (except for the `requests`'s `Session` class TBI).
* Removed dangling references to the client's `datadir` which is no longer necessary since the stateless client PR #252

This implements the client side of #187


microraiden/microraiden/client/default_http_client.py
```diff
@@ -15,42 +17,56 @@ class DefaultHTTPClient(HTTPClient):
...
             self,
             client: Client,
-            api_endpoint: str,
-            api_port: int,
             retry_interval: float = 5,
             initial_deposit: Callable[[int], int] = lambda price: 10 * price,
             topup_deposit: Callable[[int], int] = lambda price: 5 * price
     ) -> None:
-        super().__init__(client, api_endpoint, api_port)
+        HTTPClient.__init__(self, client)
```
pcppcp: Isn't `super().__init__` preferrable in P3?

devium: As far as I know, for single inheritance explicit calls are preferred. It's also more informative in my opinion. But I don't really have a preference.

pcppcp: I am also ok with either, both are ugly IMO :) But that's Python.

PR #261, opened 2017-12-08
We now have a CLA!

devium:


LefterisJP:
@devium you can also sign it by going to the CLA assistant website directly :p

Issue #256, opened 2017-12-07
Use ORM library for database management


devium:
Currently, the channel manager deals with all SQL queries manually. There are a few ORM libraries that make this a bit easier and more maintainable. Two I know of:
* SQLAlchemy https://www.sqlalchemy.org/
* peewee http://docs.peewee-orm.com/en/latest/index.html

Currently, it's not broken though, so no priority.

PR #264, opened 2017-12-11
Fixture for initializing the microraiden.requests module with a tester-aware DefaultHTTPClient

devium:
`microraiden.requests` could not be used easily in tests, since there was no fixture for it. The new `init_microraiden_requests` fixture simply calls a new variant of `microraiden.requests.init()` with the existing `default_http_client` fixture.

New µRaiden client tests can now be written like this:
```
def test_something(init_microraiden_requests, some_microraiden_server):
    import microraiden.requests
    response = microraiden.requests.get('https://localhost:5000/that_resource')
    assert response...
```


microraiden/microraiden/requests/__init__.py
```diff
@@ -3,15 +3,16 @@
...
 from microraiden.contract_proxy import ChannelContractProxy, ContractProxy
 
-client = None
-http_client = None
+_http_client = None
 
 
 def init(
+    http_client: HTTPClient = None,
+    client_: Client = None,
```
pcppcp: underscore at the end is ugly

microraiden/microraiden/requests/__init__.py
```diff
@@ -3,15 +3,16 @@
...
 from web3 import Web3
 
-from microraiden import DefaultHTTPClient, Client
+from microraiden import DefaultHTTPClient, Client, HTTPClient
 from microraiden.config import CHANNEL_MANAGER_ADDRESS, CONTRACT_METADATA
 from microraiden.contract_proxy import ChannelContractProxy, ContractProxy
 
-client = None
-http_client = None
+_http_client = None
```
pcppcp: So it's not possible to use two different keys in one session? It's probably ok to have a global now, but the library should also support creating multiple `request` contexts.

devium: The use of multiple clients/keys is planned for #256 and will follow the design of `requests.Session`.
Currently if you want multiple clients you just have to set up your own `DefaultHTTPClient`s again which is pretty much the same as calling `init()` on the global client.

microraiden/microraiden/requests/__init__.py
devium: Agreed, will fix. Looks like an earlier version.

Issue #270, opened 2017-12-11
Refactor of contract_proxy.py


devium:
We're stuck with `contract_proxy.py` until we can switch to a different library than `web3` or `pyethereum` for that matter.

Until then, we will need contract functions as defined in `contract_proxy.py` but the current design is very unsuitable for reuse in other projects. For starters, a specific contract proxy is unnecessarily bound to a specific private key, contract, and gas settings.

Instead I suggest turning most of the functionality present in `contract_proxy` into generalized static methods that work with `web3`'s `Contract` class, and that additional information like gas limits are passed with each call, with reasonable defaults (e.g. the `GAS_PRICE` and `GAS_LIMIT` constants defined in `config.py` because why not).

Functions like `create_signed_transaction` should take a private key on call.

Log queries can be generalized and don't require a function per event, essentially making the existing `get_logs()` the new entry point to log queries.

Generally, most of the `ContractProxy`'s functionality doesn't require a private key so it makes no sense for it to require one on creation.

A separate `wait_for_transaction(web3: Web3, tx_hash: str)` could also be introduced that can be monkeypatched during tests. This function is definitely needed for projects based on µRaiden.

On a more general note, we should combine functions from this module and `crypto.py` into a separate package since it's mostly functionality that should not be provided by µRaiden but rather external libraries (which we are currently lacking).

I realize this is one of our oldest classes and I wrote most of this class but good god it didn't age well.

PR #272, opened 2017-12-11
Change proxy 'content' class to Flask resource

pcppcp:
First draft. A paywalled content is now inherited `flask_restful` resource. This allows for using any http method for getting the content.
What's missing: tests, docs.

Fixes: #270 #187


microraiden/microraiden/proxy/paywalled_proxy.py
```diff
@@ -148,3 +132,14 @@ def gevent_error_handler(context, exc_info):
         if isinstance(e, ssl.SSLError) and e.reason == 'HTTP_REQUEST':
             return
         gevent.get_hub().handle_system_error(exc_info[0], exc_info[1])
+
+    def add_paywalled_resource(self, cls: Expensive, url: str, price: int, *args, **kwargs):
```
devium: `cls` would now be `Type[Expensive]` because it takes a class. Wouldn't have noticed but my IDE complained. Type hints in flask appear to be wrong. It will still complain about the `self.api.add_resource()` call though but I guess that's just because flask uses older type hints.

Also, `price` would technically be `Union[int, Callable[[str], int]]`

Sorry for being a type hint pedant.

microraiden/microraiden/test/test_proxy.py
```diff
@@ -64,10 +106,7 @@ def price_fn(url: str):
             url_to_price[url] = price
         return price
 
-    content = PaywalledContent(
-        'resource_\d', price=price_fn, get_fn=lambda url: (url.split("_")[1], 200),
-    )
-    proxy.add_content(content)
+    proxy.add_paywalled_resource(DynamicPriceResource, '/resource_<int:res_id>', price_fn)
```
devium: This is awesome. @konradkonrad will be very happy.

Issue #266, opened 2017-12-11
Follow best practices for closing_sig


loredanacirstea:
tl;dr

`closing_sig` should be the receiver's signature on:
- sender address
- channel block number
- balance
- uRaiden contract address

First proposed in https://github.com/raiden-network/microraiden/issues/52

(note, this is the current EIP712 hash implementation - will change).
```
bytes32 closing_signature_message_hash = keccak256(
          keccak256('address sender', 'uint32 block_created', 'uint192 balance', 'address contract'),
          keccak256(_sender_address, _open_block_number, _balance, address(this))
        );
```

At the moment, `closing_sig` (receiver's closing signature for cooperative close) consists only in the receiver signing the sender's signed balance proof.

Reasons for changing this:

1. There is no known vulnerability at this point due to Ethereum's ECDSA using `v` to differentiate between the possible curve points for the same signature. Nonetheless, we should follow best practices and sign actual data instead of a signature. 

2. Use EIP712, as we do in the balance proof message, for consistency. Also mentioned in the external audit of the smart contract (point 10 https://github.com/raiden-network/microraiden/issues/135)

Previous discussions on the signature logic documented in https://github.com/raiden-network/microraiden/issues/193.

To recap, current state is:
-  `balance_msg_sig` is the sender's signature of:
```
bytes32 message_hash = keccak256(
          keccak256('address receiver', 'uint32 block_created', 'uint192 balance', 'address contract'),
          keccak256(_receiver_address, _open_block_number, _balance, address(this))
        );
```
- `closing_sig` is the recever's signature on `sha3(balance_msg_sig)`

PR #273, opened 2017-12-12
Contract closing signature

loredanacirstea:
fixes https://github.com/raiden-network/microraiden/issues/266

Open to suggestions for the `message_id` values.
As suggested in https://github.com/raiden-network/microraiden/issues/134#issuecomment-347144899

tl;dr

```python
# extractBalanceProofSignature
       bytes32 message_hash = keccak256(
            keccak256(
                'string message_id',
                'address receiver',
                'uint32 block_created',
                'uint192 balance',
                'address contract'
            ),
            keccak256(
                'Sender balance proof signature',
                _receiver_address,
                _open_block_number,
                _balance,
                address(this)
            )
        );

# extractClosingSignature
    bytes32 message_hash = keccak256(
            keccak256(
                'string message_id',
                'address sender',
                'uint32 block_created',
                'uint192 balance',
                'address contract'
            ),
            keccak256(
                'Receiver closing signature',
                _sender_address,
                _open_block_number,
                _balance,
                address(this)
            )
        );

# callable by sender, receiver, delegate; closes channel immediately
function cooperativeClose(
        address _receiver_address,
        uint32 _open_block_number,
        uint192 _balance,
        bytes _balance_msg_sig,
        bytes _closing_sig)
        external
    {...}

# called just by the sender, starts the challenge period
function uncooperativeClose(
        address _receiver_address,
        uint32 _open_block_number,
        uint192 _balance)
        external
    {...}
```

- rename `verifyBalanceProof` to `extractBalanceProofSignature`, add `extractClosingSignature`
- closing signature = receiver signs channel data (sender address, block, balance, uRaiden contract address) instead of the sender's signature.
- refactor channel closing, support delegates (now, the eip712 for the closing signature really makes sense, because the receiver might use an external service to run uRaiden).
-  `cooperativeClose` (needs signatures from both sender and receiver) - can be called by sender, receiver or a delegate; the closing signature is only given when at least 1 of the parties (the receiver) wants to close the channel - so delegate closing should be safe.
   - for code simplicity, we require both of the signatures regardless of `msg.sender`
- `uncooperativeClose` - called only by the sender; no need for the sender's signature here anymore. this cannot be done by a delegate (no reason for delegates anyway)

Current gas cost (populus):
- cooperativeClose: 70429
- uncooperativeClose: 53753
- settle: 40538

### DIFF
--- a/contracts/contracts/RaidenMicroTransferChannels.sol
+++ b/contracts/contracts/RaidenMicroTransferChannels.sol
@@ -213,6 +213,7 @@ contract RaidenMicroTransferChannels {
     {
         bytes32 key = getKey(msg.sender, _receiver_address, _open_block_number);
 
+        require(channels[key].open_block_number > 0);
         require(closing_requests[key].settle_block_number == 0);
         require(_balance <= channels[key].deposit);
 
--- a/contracts/contracts/RaidenMicroTransferChannels.sol
+++ b/contracts/contracts/RaidenMicroTransferChannels.sol
@@ -298,7 +298,7 @@ contract RaidenMicroTransferChannels {
         // new contract deployment with updated code.
         bytes32 message_hash = keccak256(
             keccak256(
-                'string messageID',
+                'string message_id',
                 'address receiver',
                 'uint32 block_created',
                 'uint192 balance',
@@ -343,7 +343,7 @@ contract RaidenMicroTransferChannels {
         // new contract deployment with updated code.
         bytes32 message_hash = keccak256(
             keccak256(
-                'string messageID',
+                'string message_id',
                 'address sender',
                 'uint32 block_created',
                 'uint192 balance',
--- a/contracts/contracts/RaidenMicroTransferChannels.sol
+++ b/contracts/contracts/RaidenMicroTransferChannels.sol
@@ -189,10 +189,10 @@ contract RaidenMicroTransferChannels {
         external
     {
         // Derive sender address from signed balance proof
-        address sender = verifyBalanceProof(_receiver_address, _open_block_number, _balance, _balance_msg_sig);
+        address sender = extractBalanceProofSignature(_receiver_address, _open_block_number, _balance, _balance_msg_sig);
 
         // Derive receiver address from closing signature
-        address receiver = verifyClosingSignature(sender, _open_block_number, _balance, _closing_sig);
+        address receiver = extractClosingSignature(sender, _open_block_number, _balance, _closing_sig);
         require(receiver == _receiver_address);
 
         // Both signatures have been verified and the channel can be settled.
@@ -281,7 +281,7 @@ contract RaidenMicroTransferChannels {
     /// @param _balance The amount of tokens owed by the sender to the receiver.
     /// @param _balance_msg_sig The balance message signed by the sender.
     /// @return Address of the balance proof signer.
-    function verifyBalanceProof(
+    function extractBalanceProofSignature(
         address _receiver_address,
         uint32 _open_block_number,
         uint192 _balance,
@@ -326,7 +326,7 @@ contract RaidenMicroTransferChannels {
     /// @param _balance The amount of tokens owed by the sender to the receiver.
     /// @param _closing_sig The receiver's signed balance message, containing the sender's address.
     /// @return Address of the closing signature signer.
-    function verifyClosingSignature(
+    function extractClosingSignature(
         address _sender_address,
         uint32 _open_block_number,
         uint192 _balance,
--- a/contracts/contracts/RaidenMicroTransferChannels.sol
+++ b/contracts/contracts/RaidenMicroTransferChannels.sol
@@ -172,38 +172,14 @@ contract RaidenMicroTransferChannels {
         require(token.transferFrom(msg.sender, address(this), _added_deposit));
     }
 
-    /// @notice Function called when any of the parties wants to close the channel and settle.
-    /// Receiver needs a balance proof to immediately settle, sender triggers a challenge period.
+    /// @notice Function called by the sender, receiver or a delegate, with all the needed
+    /// signatures to close the channel and settle immediately.
     /// @param _receiver_address The address that receives tokens.
     /// @param _open_block_number The block number at which a channel between the
     /// sender and receiver was created.
     /// @param _balance The amount of tokens owed by the sender to the receiver.
     /// @param _balance_msg_sig The balance message signed by the sender.
-    function uncooperativeClose(
-        address _receiver_address,
-        uint32 _open_block_number,
-        uint192 _balance,
-        bytes _balance_msg_sig)
-        external
-    {
-        address sender = verifyBalanceProof(_receiver_address, _open_block_number, _balance, _balance_msg_sig);
-
-        if(msg.sender == _receiver_address) {
-            settleChannel(sender, _receiver_address, _open_block_number, _balance);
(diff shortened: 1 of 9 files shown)


contracts/tests/fixtures.py
```diff
@@ -2,8 +2,12 @@
 from utils.logs import LogHandler
 
 
+print_the_logs = False
+
+
 MAX_UINT = 2 ** 256 - 1
```
pcppcp: I think `MAX_UINT256` would be more consistent here.

contracts/contracts/RaidenMicroTransferChannels.sol
```diff
@@ -295,15 +297,72 @@ contract RaidenMicroTransferChannels {
         // ! Note that EIP712 might change how hashing is done, triggering a
         // new contract deployment with updated code.
         bytes32 message_hash = keccak256(
-          keccak256('address receiver', 'uint32 block_created', 'uint192 balance', 'address contract'),
-          keccak256(_receiver_address, _open_block_number, _balance, address(this))
+            keccak256(
+                'string messageID',
```
LefterisJP: For consistency's sake can you snake case it? `message_id`?

contracts/contracts/RaidenMicroTransferChannels.sol
```diff
@@ -295,15 +297,72 @@ contract RaidenMicroTransferChannels {
...
+    {
+        // The variable names from below will be shown to the sender when signing
+        // the balance proof, so they have to be kept in sync with the Dapp client.
+        // The hashed strings should be kept in sync with this function's parameters
+        // (variable names and types).
+        // ! Note that EIP712 might change how hashing is done, triggering a
+        // new contract deployment with updated code.
+        bytes32 message_hash = keccak256(
+            keccak256(
+                'string messageID',
```
LefterisJP: change as suggested above

contracts/contracts/RaidenMicroTransferChannels.sol
```diff
@@ -295,15 +297,72 @@ contract RaidenMicroTransferChannels {
...
 
+    /// @dev Returns the receiver address extracted from the closing signature.
+    /// Works with eth_signTypedData https://github.com/ethereum/EIPs/pull/712.
+    /// @param _sender_address The address that sends tokens.
+    /// @param _open_block_number The block number at which a channel between the
+    /// sender and receiver was created.
+    /// @param _balance The amount of tokens owed by the sender to the receiver.
+    /// @param _closing_sig The receiver's signed balance message, containing the sender's address.
+    /// @return Address of the closing signature signer.
+    function verifyClosingSignature(
```
LefterisJP: The function name here is misleading. Either change it to `extractClosingSignature` or add the verification check that you perform [outside](https://github.com/raiden-network/microraiden/pull/273/files#diff-f70d66082794ad8ac4eb9c1e46e17371R196) of the function inside.

contracts/tests/fixtures.py
```diff
@@ -42,6 +63,12 @@ def contract_params(request):
     return request.param
 
 
+@pytest.fixture(params=channel_values)
+def channel_params(request):
+    return request.param
+
+
+
```
LefterisJP: super nitpicky: Why 3 newlines? Just 2 should be enough.

contracts/tests/fixtures_uraiden.py
```diff
@@ -73,52 +77,57 @@ def uraiden_instance(owner, uraiden_contract):
     return uraiden_instance
 
 
-@pytest.fixture(params=['20', '223'])
-def get_channel(web3, request, owner, get_accounts, event_handler, print_gas, txn_gas, get_block):
-    def get(uraiden_instance, token_instance, deposit, sender=None, receiver=None):
-        contract_type = request.param
-        ev_handler = event_handler(uraiden_instance)
+@pytest.fixture
+def get_channel(web3, channel_params, owner, get_accounts, uraiden_instance, token_instance, event_handler, print_gas, txn_gas, get_block):
```
LefterisJP: style: Too many args. Split in multiple lines with each arg in its own line (assuming you follow the raiden styling guide)

contracts/tests/utils.py
```diff
@@ -3,9 +3,17 @@
 
 def balance_proof_hash(receiver, block, balance, contract):
     return eth_signed_typed_data_message(
-        ('address', ('uint', 32), ('uint', 192), 'address'),
-        ('receiver', 'block_created', 'balance', 'contract'),
-        (receiver, block, balance, contract)
+        ('string', 'address', ('uint', 32), ('uint', 192), 'address'),
+        ('messageID', 'receiver', 'block_created', 'balance', 'contract'),
```
LefterisJP: As suggested above change messageID field name also in this file

contracts/contracts/RaidenMicroTransferChannels.sol
loredanacirstea: The reason was name consistency with `verifyBalanceProof` which does the same thing.
So, in this case I should also change `verifyBalanceProof` to `extractBalanceProofSignature`, right?

contracts/tests/fixtures_uraiden.py
loredanacirstea: Yes, only ran flake8 on test_channel_close.py at the time... But I refactored this part in https://github.com/raiden-network/microraiden/pull/310 . I would leave it like this for now, to have less merging conflicts.

contracts/contracts/RaidenMicroTransferChannels.sol
LefterisJP: I would say so yes

loredanacirstea:
@LefterisJP , you can approve if the changes are satisfactory. Then, the PR can be merged to master. uRaiden works by using the contract data from https://github.com/raiden-network/microraiden/blob/d7aff7168496a5e0f4354590dcb50013a33b33d5/microraiden/microraiden/data/contracts.json.
We will have another PR for changing the contract data & proxy logic.

contracts/contracts/RaidenMicroTransferChannels.sol
```diff
@@ -172,57 +172,59 @@ contract RaidenMicroTransferChannels {
...
+        address sender = extractBalanceProofSignature(_receiver_address, _open_block_number, _balance, _balance_msg_sig);
 
-        if(msg.sender == _receiver_address) {
-            settleChannel(sender, _receiver_address, _open_block_number, _balance);
-        } else {
-            require(msg.sender == sender);
-            initChallengePeriod(_receiver_address, _open_block_number, _balance);
-        }
+        // Derive receiver address from closing signature
+        address receiver = extractClosingSignature(sender, _open_block_number, _balance, _closing_sig);
```
jannikluhn: Is it possible for A and B to open channels with each other in both directions? If so, the `_balance_msg_sig` of one channel seems to be reusable as `_closing_sig` of the other.

It's a bit artificial, because in practice there won't be two channels between two parties opened in the same block, but maybe still worth considering.

loredanacirstea: The signatures now differ by the `message_id` value and `address receiver` vs. `address sender` in the message hash. So you know that one is a balance proof and one is a closing signature and cannot swap them.

jannikluhn: Oops, yeah, you're right. Sorry, then it's fine.

PR #274, opened 2017-12-12
Update requirements.txt

pcppcp:
ethereum-utils library refuses to start in version >0.6.0 and
ethereum-abi-utils also raises an exception in version >0.4.0

The reason for this is rename of the library to eth-utils and
eth-abi-utils, respectively.

Because these two are dependencies of other libraries, we must
freeze versions until it's fixed upstream.


LefterisJP:
Looks good to me. When tests pass we can merge

PR #275, opened 2017-12-12
Refactoring of contract_proxy.py methods

devium:
This addresses issue #270 

- All functionality that used to be provided by `contract_proxy.py` is now made stateless. Previous class methods are now global methods available in `microraiden.utils.contract`, dynamically taking former contexts like private keys, `Web3`, or `Contract` objects. Instead of a `ContractProxy` object, modules are now expected to keep a `Contract` (from `web3.py`) object in their state.
- A lot of the unnecessary contract functions that used to be exposed in `contract_proxy.py` like `get_channel_created_logs()` must now be executed using `get_logs()`, which is just as easy but not as messy with respect to reuse.
- `crypto.py`, `contract.py`, and private key management have been moved to the `microraiden.utils` package and can directly be imported from `microraiden.utils`.
- Removed a cyclic dependency between clients and channels, by moving shared state to a new `Core` class (name pending) and adding an `on_settle` callback.
- Contract/on-chain functionality is no longer tester-aware but exposes relevant functions that are easily patched to be tester-compatible.
- A new `wait_for_transaction()` function. Tested but currently unused.

Labelled WIP because there's probably a huge rebase merge coming up.


microraiden/microraiden/client/core.py
```diff
@@ -0,0 +1,28 @@
+from web3 import Web3
+
+from microraiden.config import CONTRACT_METADATA, TOKEN_ABI_NAME, CHANNEL_MANAGER_ABI_NAME
+from microraiden.utils import privkey_to_addr
+
+
+class Core():
```
pcppcp: `ClientContracts` maybe?

devium:
I went with `Context`. It's almost as generic but a tad more descriptive. `ChainContext` or `ETHContext` or something would maybe work too.

PR #282, opened 2017-12-13
Proxy: return resource if its price is 0

pcppcp:


microraiden/microraiden/test/test_proxy.py
```diff
@@ -8,6 +8,9 @@
 from microraiden.proxy.resources import Expensive
 from microraiden.proxy.paywalled_proxy import PaywalledProxy
 
+import logging
```
devium: What's this for?

pcppcp: I had some logging enabled there, I'd leave it there, we'll soon need logs to investigate slow proxy tests.

PR #285, opened 2017-12-13
Faster ETH ticker test

devium:
The ETH ticker test now mocks the bitfinex API call and allows for faster polling times.

The test might in some cases be too short for the ticker window to appear. It should at most take 2 seconds.

### DIFF
--- a/microraiden/microraiden/test/test_eth_ticker.py
+++ b/microraiden/microraiden/test/test_eth_ticker.py
@@ -1,8 +1,11 @@
 import pytest  # noqa: F401
+from _pytest.monkeypatch import MonkeyPatch
+from flask import jsonify
 
 from microraiden import DefaultHTTPClient
 from microraiden.examples.eth_ticker import ETHTickerClient, ETHTickerProxy
 from microraiden.proxy.paywalled_proxy import PaywalledProxy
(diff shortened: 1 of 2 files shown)


microraiden/microraiden/test/test_eth_ticker.py
```diff
@@ -11,20 +14,30 @@ def test_eth_ticker(
...
-    ticker = ETHTickerClient(sender_privkey, httpclient=default_http_client)
+    def get_patched(*args, **kwargs):
+        body = {
+            'mid': '682.435', 'bid': '682.18', 'ask': '682.69', 'last_price': '683.16',
+            'low': '532.97', 'high': '684.0', 'volume': '724266.25906224',
+            'timestamp': '1513167820.721733'
+        }
+        return jsonify(body)
+
+    monkeypatch.setattr(PaywalledProxyUrl, 'get', get_patched)
```
pcppcp: This will work, but please double check if the method is replaced for this test only and we can use the class normally in other tests.

devium: The `monkeypatch` fixture takes care of that. That's the beauty of it.

pcppcp: cool, let's merge this then

Issue #259, opened 2017-12-07
Handle HTTPClient channel closure in case the server does not hand out a closing signature


devium:
Currently, if the server rejects the request for a closing signature, the `HTTPClient` simply logs an error and does nothing. Not good.

In this case, the client should initiate an uncooperative close.

PR #292, opened 2017-12-14
Client handles nonexisting channel message

devium:
Issue #263

PR #293, opened 2017-12-14
Client handles denial of closing signature

devium:
It tries to close the channel with a balance of 0, although this behavior is easily overridden using the new `on_cooperative_close_denied` callback. #259


microraiden/microraiden/test/test_http_client.py
```diff
@@ -536,3 +539,49 @@ def test_requests(
...
+    with requests_mock.mock() as server_mock:
+        headers = {
+            HTTPHeaders.TOKEN_ADDRESS: token_address,
+            HTTPHeaders.CONTRACT_ADDRESS: channel_manager_address,
+            HTTPHeaders.RECEIVER_ADDRESS: receiver_address,
+            HTTPHeaders.PRICE: '3'
+        }
+        headers = [headers.copy() for i in range(2)]
+
+        i = 0
```
pcppcp: lolwut?
```
i = 0
i += 1
```

microraiden/microraiden/client/http_client.py
```diff
@@ -178,6 +177,10 @@ def on_invalid_contract_address(
     def on_payment_requested(self, method: str, url: str, response: Response, **kwargs):
         return True
 
+    def on_cooperative_close_denied(self, endpoint_url: str, channel: Channel, response: Response):
+        log.warning('No closing signature received. Closing noncooperatively on a balance of 0.')
+        channel.close(0)
```
pcppcp: That's mean (:

microraiden/microraiden/test/test_http_client.py
```diff
@@ -536,3 +539,49 @@ def test_requests(
...
+
+        url = 'http://{}/something'.format(api_endpoint_address)
+        server_mock.get(url, [
+            {'status_code': 402, 'headers': headers[0]},
+            {'status_code': 200, 'headers': headers[1], 'text': 'success'},
+        ])
+        channel_url = re.compile(
+            'http://{}/api/1/channels/0x.{{40}}/\d+'.format(api_endpoint_address)
+        )
+        server_mock.delete(channel_url, [
```
pcppcp: Just a note: atm we have two `DELETE` methods implemented. One for `/channels/<sender_addr>/<opening_block>` and the other for `/channels/<sender_addr>`

We should make them consistent - one returns text string, the other JSON.

microraiden/microraiden/test/test_http_client.py
```diff
@@ -536,3 +539,49 @@ def test_requests(
...
+    ).start()
+
+    with requests_mock.mock() as server_mock:
+        headers = {
+            HTTPHeaders.TOKEN_ADDRESS: token_address,
+            HTTPHeaders.CONTRACT_ADDRESS: channel_manager_address,
+            HTTPHeaders.RECEIVER_ADDRESS: receiver_address,
+            HTTPHeaders.PRICE: '3'
+        }
+        headers = [headers.copy() for i in range(2)]
```
pcppcp: why not just
`headers = [ headers.copy(), headers.copy() ]`

microraiden/microraiden/test/test_http_client.py
devium: Both this and the `range(2)` magic up there is for easy reordering. You can just take one header configuration and move the entire block somewhere else in the header chain within the test. I often use this instead of explicit indices for just this reason.

If it's too confusing I can change it.

microraiden/microraiden/test/test_http_client.py
devium: Which one does `/channels/<sender_addr>` default to if there are multiple channels open?

microraiden/microraiden/test/test_http_client.py
devium: Might make more sense in the other PR:
https://github.com/raiden-network/microraiden/pull/292/files#diff-6909d3c2a4b37b2ee0887ccc30f8107cR577

microraiden/microraiden/test/test_http_client.py
pcppcp: it requires `{'open_block' : ..., 'signature': ... }` json as a request data

microraiden/microraiden/test/test_http_client.py
pcppcp: I see. I still think it's much more readable if you just index directly without using variable. In test it doesn't matter that much of course.

microraiden/microraiden/test/test_http_client.py
pcppcp: or 
```
[headers.copy() for _ in range(2)]
```
(that'd make sense in the other PR you've linked)

Issue #265, opened 2017-12-11
Add `microraiden.requests.Session` class, analogous to `requests.Session`


devium:
The new `microraiden.requests` module offers easy access to `requests`-like request functions but doesn't expose an equivalent to the `requests.Session` class that provides some parameter and connection persistence over multiple requests.

See http://docs.python-requests.org/en/master/user/advanced/ for more details on what the `requests.Session` class supports.

It probably makes sense to adopt the `requests` architecture, meaning that `HTTPClient` would inherit `requests.Session` and use the underlying session for all of its requests.

An `HTTPClient` would then again only manage a single channel, while the `microraiden.requests.request` and similar functions create a short-lived `HTTPClient` (might make sense to rename this to `Session` too) object based on global default settings set in `microraiden.requests.init()`, similar to how it's done in `requests`:
https://github.com/requests/requests/blob/22d12b0501fa633b80bcda303a718696e408ebfb/requests/api.py#L16-L58

devium:
It makes sense to also merge `DefaultHTTPClient` and `HTTPClient` into a single class, since single parts of the `DefaultHTTPClient` logic can still easily be overridden by overriding the `on_` callbacks. `HTTPClient` then contains the reasonable default logic.

Let's just call the resulting class `Session` to keep it consistent with the `requests` module.

PR #295, opened 2017-12-14
requests-like Session class

devium:
Addresses issue #265 
~Is based on both PR #292 and PR #293~ Rebased on merged master.

The new `Session` class combines functionality of the former classes `HTTPClient` and `DefaultHTTPClient` with `requests.Session`.

`Session` directly inherits `requests.Session` and overrides its methods for HTTP requests.

Each `Session` object only manages a single channel and tries to stick with until it is closed. By setting the `close_channel_on_exit` flag in the constructor, the channel is automatically closed when the underlying `requests.Session` is closed.

Accordingly, `microraiden.requests.Session` can be used in a `with` statement but doesn't close its channel per default unless `close_channel_on_exit` is set to `True`.

The global methods in `microraiden.requests` create a temporary session with the passed parameters. Keep in mind though that these methods require additional arguments to initialize the `Session`'s `Client`. At minimum, this requires a private key.


microraiden/microraiden/client/client.py
```diff
@@ -28,11 +28,10 @@ def __init__(
             channel_manager_address: str = CHANNEL_MANAGER_ADDRESS,
             web3: Web3 = None
     ) -> None:
-        assert private_key or key_path
-        assert not private_key or isinstance(private_key, str)
+        assert private_key is not None or key_path is not None
```
pcppcp: I think having just a single variable `private_key` and use a regex to detect hex-string and fallback to `os.path.isfile` if the regex test fails would make it much simpler. Maybe having an absolute path as a requirement would make it even more secure.

microraiden/microraiden/client/client.py
```diff
@@ -28,11 +28,10 @@ def __init__(
...
 
         # Load private key from file if none is specified on command line.
-        if not private_key:
+        if private_key is None:
             private_key = get_private_key(key_path, key_password_path)
             assert private_key is not None
```
pcppcp: An exception should be raised if the key isn't valid.

microraiden/microraiden/client/client.py
devium: And another client constructor argument has been eliminated! Nice.

microraiden/microraiden/requests/__init__.py
```diff
@@ -1,68 +1,46 @@
...
-    assert _http_client, 'microraiden.requests has not been initialized. ' \
-                         'Please call microraiden.requests.init() first.'
+
+from microraiden import Session, Client
 
 
 def request(method: str, url: str, **kwargs) -> Response:
-    _check_init()
-    return _http_client.request(method, url, **kwargs)
+    session_kwargs = {
```
devium: A note since this question already came up:

`requests` is strict on its `kwarg`s. If it doesn't know any of them, it raises an exception. So we have to filter out arguments that are destined for `Session` or `Client`. This is why I'm using `inspect` here.

Alternatively, one could use different `cfg` dicts for `Session` and `Client` which would result in a slightly different usage of these global methods. It's a matter of taste. `inspect` and shared `kwargs` vs separate `client_cfg` and `session_cfg` arguments.

Would love to hear your input.

pcppcp: I really like the `inspect` method. It's very clean. I'll probably use the same for proxy's `Expensive` resources.

PR #296, opened 2017-12-15
Make uRaiden available on pypi

### LINKED ISSUE #217: Make microraiden available on Python Package Index
@konradkonrad suggested that making uRaiden available on pip would make the integration much easier for _other projects_.

pcppcp:
Fixes #217

### DIFF
--- a/microraiden/setup.py
+++ b/microraiden/setup.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 
 try:
-    from setuptools import setup
+    from setuptools import setup, find_packages
 except ImportError:
     from distutils.core import setup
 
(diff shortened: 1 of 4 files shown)


microraiden/setup.py
```diff
@@ -10,10 +10,84 @@
...
+    'install_requires': reqs,
+    'packages': find_packages(exclude=['test']),
+    'package_data': {'microraiden': ['data/contracts.json',
+                                     'webui/js/*',
+                                     'webui/index.html',
+                                     'webui/microraiden/dist/umd/microraiden.js']},
+    'classifiers': [
+        'Development Status :: 3 - Alpha',
+        'Intended Audience :: Developers',
+        'License :: OSI Approved :: BSD License',
```
palango: I think we should use MIT license here (even though MIT and BSD 2-clause are equivalent)

pcppcp: It's what I've copypasted from raiden, maybe it should be changed there too.

palango: Yeah, just did that in my branch

PR #297, opened 2017-12-16
fix TOC.

sot528:


loredanacirstea:
Thank you for your submission, but we are moving to .rst format (https://github.com/raiden-network/microraiden/pull/345) and TOC links are not needed anymore.

PR #299, opened 2017-12-18
Update contract compiled assets for ver 0.1.0

loredanacirstea:
Contract assets were outdated


loredanacirstea:
@pcppcp , I think tests are failing because the contract now requires:
```
require(_challenge_period >= 500);
```
and tests use 30: 
https://github.com/raiden-network/microraiden/blob/d7aff7168496a5e0f4354590dcb50013a33b33d5/microraiden/microraiden/test/fixtures/web3.py#L84

pcppcp:
Can you try to adjust it in the tests? I think there's no test that requires a short period, we can use 500 for everything.

PR #302, opened 2017-12-18
Add MacOS build requirements

gleim:


README.md
```diff
@@ -86,6 +86,22 @@ virtualenv -p python3 env
 pip install -e microraiden
 ```
 
+* if necessary to complete the build, install latest OpenSSL on MacOS
```
loredanacirstea: Please add this after `Quick Start`:
### MacOS build issues with OpenSSL
<your suggestions>

loredanacirstea:
@gleim , can you restrict the PR to only the Readme addition so it can be merged?

README.md
gleim: I think this is what you're looking for, let me know if I can provide further improvement.

PR #304, opened 2017-12-18
Update eth tester dependency

gleim:


microraiden/requirements-dev.txt
```diff
@@ -1,5 +1,5 @@
 -r requirements.txt
 pytest
-ethereum-tester
+eth-tester
```
pcppcp: Please revert this change. There are some compatibility issues in the ethereum libraries we are using and we need to fix these before we can use new `eth-tester` and others.

pcppcp:
I think you can close this PR and remove `requirements-dev.txt` changes in #302 
`eth-tester` compatibility is hotfixed by freezing the version (PR #306)

PR #305, opened 2017-12-19
Bug fix: proxy resource headers take precedence over default headers

devium:
This came up in a µRaiden application: When the returned resource had an explicit content type (for example `application/json` from a `jsonify()` call) the respective header wasn't merged properly but simply extended. This resulted in headers like `'Content-Type': 'application/json,application/json'` or `'Content-Type': 'application/json,text'`.

In this PR, header fields returned by the resource take precedence over default headers. That includes possibly overriding `RDN-...` fields. Not sure one would want that but just note that it's possible now.

### DIFF
--- a/microraiden/microraiden/proxy/resources/paywall_decorator.py
+++ b/microraiden/microraiden/proxy/resources/paywall_decorator.py
@@ -121,12 +121,26 @@ class Paywall(object):
 
         # all ok, return actual content
         resp = method(request.path, *args, **kwargs)
+
+        # merge headers, resource headers take precedence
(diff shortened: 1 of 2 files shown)


andrevmatos:
For reference, I also hit this issue, and had a fix staged: https://github.com/andrevmatos/microraiden/commit/27404042cd709ffadc1d310ac2d763d5fbfadfe6#diff-17130143bd0e4ceb6428aede784d88b0R125

pcppcp:
Good catch. I didn't know that the valid content-type header can contain multiple types.

PR #306, opened 2017-12-19
Freeze ethereum-utils==0.4.0

pcppcp:
Until we have full populus-2.2.0 compatibility we must use older
ethereum-utils version.


loredanacirstea:
Fixed in https://github.com/raiden-network/microraiden/pull/327 by upgrading to populus 2.2.0 and `eth-utils`

PR #309, opened 2017-12-19
Update README.md

seangeleno:
Was trying to read more about standard and saw they have dedicated repo,  thought it would be a more explicit reference.


loredanacirstea:
@seangeleno , thank you for your effort. 
However, a reference to https://github.com/ethereum/EIPs is preferred as documentation for any implemented or proposed Ethereum standard. That is where all public discussions take place and where other sources for code or additional docs are linked.
This is why I will have to close your PR.

PR #310, opened 2017-12-20
Contract tests refactor open

loredanacirstea:
To be reviewed after https://github.com/raiden-network/microraiden/pull/273 is merged.

- move create channel tests from the `get_channel` fixture used in all tests to dedicated tests
- shortens travis time relative to the above PR

### DIFF
--- a/contracts/tests/test_channel_create.py
+++ b/contracts/tests/test_channel_create.py
@@ -123,30 +123,39 @@ def test_channel_erc223_create_bounty_limit(
 
 
 def test_create_token_fallback_uint_conversion(
-    contract_params,
-    owner,
-    get_accounts,
-    uraiden_instance,
-    token_instance):
-    token = token_instance
+        owner,
+        get_accounts,
+        uraiden_contract,
+        get_token_contract):
+    token = get_token_contract([MAX_UINT192 + 100, 'CustomToken', 'TKN', 18])
+    uraiden = uraiden_contract(token)
     (sender, receiver) = get_accounts(2)
-
-    # Make sure you have a fixture with a supply > 2 ** 192 + 100
-    deposit = contract_params['supply'] - 100
     txdata = bytes.fromhex(receiver[2:].zfill(40))
 
     # Fund accounts with tokens
-    token.transact({"from": owner}).transfer(sender, deposit)
-    assert token.call().balanceOf(sender) == deposit
+    token.transact({"from": owner}).transfer(sender, MAX_UINT192 + 5)
+    assert token.call().balanceOf(sender) == MAX_UINT192 + 5
 
     # Open a channel with tokenFallback
-    if deposit > 2 ** 192:
-        with pytest.raises(tester.TransactionFailed):
-            txn_hash = token_instance.transact({"from": sender}).transfer(
-                uraiden_instance.address,
-                deposit,
-                txdata
-            )
+    # uint192 deposit = uint192(_deposit), where _deposit is uint256
+    with pytest.raises(tester.TransactionFailed):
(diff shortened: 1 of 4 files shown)


contracts/tests/test_channel_create.py
```diff
@@ -115,6 +117,56 @@ def test_channel_223_create_bounty_limit(get_block, owner, get_accounts, uraiden
...
+def test_create_token_fallback_uint_conversion(
+    contract_params,
+    owner,
+    get_accounts,
+    uraiden_instance,
+    token_instance):
+    token = token_instance
+    (sender, receiver) = get_accounts(2)
+
+    # Make sure you have a fixture with a supply > 2 ** 192 + 100
```
LefterisJP: Shouldn't you put an `assert` instead of a comment here?

contracts/tests/test_channel_create.py
```diff
@@ -115,6 +117,56 @@ def test_channel_223_create_bounty_limit(get_block, owner, get_accounts, uraiden
...
+    if deposit > 2 ** 192:
+        with pytest.raises(tester.TransactionFailed):
+            txn_hash = token_instance.transact({"from": sender}).transfer(
+                uraiden_instance.address,
+                deposit,
+                txdata
+            )
+
+
+def test_channel_223_event(owner, get_accounts, uraiden_instance, token_instance, event_handler, print_gas):
```
LefterisJP: what is channel 223? Is it ERC223? Then please specify.

contracts/tests/test_channel_create.py
```diff
@@ -115,6 +117,56 @@ def test_channel_223_create_bounty_limit(get_block, owner, get_accounts, uraiden
...
+
+    # Check creation event
+    ev_handler.add(
+        txn_hash,
+        uraiden_events['created'],
+        checkCreatedEvent(sender, receiver, deposit)
+    )
+    ev_handler.check()
+
+    print_gas(txn_hash, 'channel_223_create')
```
LefterisJP: the print gas is for test debugging purposes?

contracts/tests/test_channel_create.py
```diff
@@ -190,28 +242,80 @@ def test_channel_20_create_bounty_limit(owner, get_accounts, uraiden_instance, t
     assert channel_data[1] == channel_deposit_bugbounty_limit
 
 
-def test_create_token_fallback_uint_conversion(
-    contract_params,
-    owner,
-    get_accounts,
-    uraiden_instance,
-    token_instance):
+def test_channel_20_event(owner, get_accounts, uraiden_instance, token_instance, event_handler, txn_gas, print_gas):
```
LefterisJP: what is channel 20 event? Is t ERC20 then please specify :P

contracts/tests/test_channel_create.py
loredanacirstea: Yes, quick way to have some values after testing.

contracts/tests/test_channel_create.py
loredanacirstea: I was using an array of contract parameters to run the tests multiple times. But this is an edge case, so I will handle it only in this test with clear parameters. Refactor on the way.

loredanacirstea:
Refactored the uint256 -> uint192 test , but I left one positive test commented due the deposit bug bounty limit. I'll open an issue to not forget about it. The tests above it are more important, so it doesn't make sense now to find complicated ways to add this one.

Also added a TODO in the code:
https://github.com/loredanacirstea/microraiden/blob/3b0e7f1c8c207b08a0e2214ed04fe842b238725d/contracts/tests/test_channel_create.py#L154

PR #311, opened 2017-12-20
Contract update requirements

loredanacirstea:
fixes https://github.com/raiden-network/microraiden/issues/203

Update to:
```
populus>=2.2.0
web3>=v4.0.0-beta.4
```
- Rename populus configuration file to project.json
- Replace old web3 log formatters in contract tests (note: the log filter class should be replaced altogether with https://github.com/ethereum/web3.py/blob/473cdb8ea803da908b9d14d46fece75c324211b9/docs/filters.rst#event-log-filters)
- Hex encode bytearray input in contract tests
- Use eth_utils._same_address for comparing addresses (normalized and checksummed representations)

PR #313, opened 2017-12-20
Compatibility fix for Populus>2.0

pcppcp:


microraiden/microraiden/channel_manager/blockchain.py
```diff
@@ -138,7 +138,7 @@ def _update(self):
             **filters_unconfirmed
         )
         for log in logs:
-            assert log['args']['_receiver'] == self.cm.state.receiver
+            assert log['args']['_receiver'].lower() == self.cm.state.receiver
```
loredanacirstea: Suggestion: use `is_same_address` from `eth_utils` or `assert log['args']['_receiver'].lower() == self.cm.state.receiver.lower()` in case `cm.state.receiver` is not normalized (future-proofed)

pcppcp: thanks, it definitely needs some polishing.
BTW I've used your `LogFilter`, definitely helped a lot!

pcppcp: when everything is working I'll merge your PR #311 to this, do you agree?

loredanacirstea: yes, it is better to have it in one place.

loredanacirstea:
43 min for travis to run the contract tests with `web3==3.16.3 `.

Locally, just by changing the web3 version I get:
`web3==3.16.3` - 564.42 sec -> 9.407 min
`web3==4.0.0-beta.4` - 170.51 sec -> 2.8 min

total time running the contract tests. both times with `-n 4` (4 cpus)

Possible explanation: 
- populus dropped gevent support from `2.0.0` 
- web3.py - "All filter calls are synchronous, gevent integration dropped" from `v4.0.0-beta.1`. 
So something is not compatible when you try populus `2.x` with web3.py < `4.x`

I would actually be in favor of using `v4.0.0-beta.4`. Aside from being in `beta`, were there other reasons not to try using it?

pcppcp:
Hm, it depends what _beta_ means here. We can use it and hope they'll make it stable soon, or support both v4 and v3.
My main concern was that the version I've tried to used didn't have dependencies set right (using 'old' `ethereum-*` libraries that refuse to work now).
Maybe let's have the beta now and I'll try to investigate what the problem with test slowdown really is if there'll be some time for that.

pcppcp:
But the current state where it takes 42 minutes to finish the CI would be really slowing the developement.. Let's switch to beta. We can figure out monkeypatch stuff later.

loredanacirstea:
Continued in https://github.com/raiden-network/microraiden/pull/327

PR #320, opened 2017-12-31
Infura support, logging tweaks

PiotrTrzpil:
1. `web3.eth.defaultAccount = privkey_to_addr(private_key)` - this is required in case of Infura, because otherwise web3 tries to request `coinbase` for some RPC calls to set a value for `from` field. If `defaultAccount` is set, then it uses that.
2. ~~`to_block=current_block + 100` - if this is missing, then web3 is setting `"pending"` as `toBlock` which Infura does not seem to like (returns empty results). Another way would be to force it somehow to not set this field, but I couldn't find how.~~
3. Some logging tweaks that came very handy while getting it to work. (Should be in a separate PR perhaps?)

### DIFF
--- a/microraiden/microraiden/client/channel.py
+++ b/microraiden/microraiden/client/channel.py
@@ -126,7 +126,7 @@ class Channel:
             self.deposit += deposit
             return event
         else:
-            log.error('No event received.')
+            log.error('No ChannelToppedUp event received.')
             return None
 
     def close(self, balance=None):
@@ -176,7 +176,7 @@ class Channel:
             self.state = Channel.State.settling
             return event
         else:
-            log.error('No event received.')
+            log.error('No ChannelCloseRequested event received.')
             return None
 
     def close_cooperatively(self, closing_sig: bytes):
@@ -234,7 +234,7 @@ class Channel:
             self.state = Channel.State.closed
             return event
         else:
-            log.error('No event received.')
+            log.error('No ChannelSettled event received.')
             return None
 
     def settle(self):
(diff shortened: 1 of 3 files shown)


LefterisJP:
Hello @PiotrTrzpil and thank you for the Pull Request. Happy new year too! :)

Regarding `2` it feels like a hack and a problem with infura. Why would it complain for `toBlock` being set to pending?

All the rest seems fine to me and yeah the logging stuff could have been in another PR but it's fine here too.

PiotrTrzpil:
@LefterisJP Thanks, you too! Great work you are doing here.

Yeah that `to_block=current_block + 100` was a false path, I could not reproduce it now - removed it.

The setup i'm using (Infura on Rinkeby with `eth_ticker`)  seems to work less reliably now, as I'm getting `Channel not registered by server. Retrying in 5 seconds.` that does not recover, around 50%, and not once succeeded in settling the channel when exiting.

 I'll try to investigate more later.

andrevmatos:
@PiotrTrzpil we experienced similar issues on the JS client on Kovan due to a lot of MetaMask (Infura?) nodes being outdated and not syncing, not getting events and even returning bad current block numbers. That's why we switched our main examples to Ropsten (MM/infura Ropsten nodes are in way better shape). If you want, you may test on Ropsten to confirm this is actually the issue. After, if you want, rebase your PR, please. Thank you for it.

PiotrTrzpil:
Looks like we'll have no luck with Infura for now, as they don't support e.g. eth_newFilter (https://github.com/INFURA/infura/issues/29#issuecomment-336589492) which is now used when creating `LogFilter`. Hence, I'll close the PR.

Issue #147, opened 2017-11-15
Open/topUp by delegates


loredanacirstea:
A channel for Bob is paid for by a delegate with `createChannel` and `topUp`. Channel owner is Bob.

Idea: add trusted contracts to RaidenMicroTransferChannels

PR #322, opened 2018-01-02
Open/topUp by delegate (trusted) contracts

loredanacirstea:
closes https://github.com/raiden-network/microraiden/issues/147

- `trusted_contracts` array - passed as a contract constructor argument.
- add `addTrustedContracts` only callable by `owner_address`
  - gas used for adding 1 trusted contract: `46888`
  - gas used for adding 3 trusted contracts: `97644`
- add `removeTrustedContracts` only callable by `owner_address`
  - gas used for adding 1 trusted contract: `16232`
  - gas used for adding 3 trusted contracts: `20964`
- add event `TrustedContract(address indexed _trusted_contract_address, bool _trusted_status)`
- a trusted contract can send tokens and open & top up channels in behalf of a `sender` (support for both ERC20 & ERC223 contracts)
  - `_data` argument from `tokenFallback` now has to contain:
```py
# create channel
sender_address (20 bytes) + receiver_address (20 bytes)
# top up
sender_address (20 bytes) + receiver_address (20 bytes) + open_block_number (4 bytes)
```
  - rename ERC20 compatible functions to `createChannel` , `topUp`
  - add ERC20 compatible functions: `createChannelDelegate`, `topUpDelegate`

### DIFF
--- a/contracts/contracts/RaidenMicroTransferChannels.sol
+++ b/contracts/contracts/RaidenMicroTransferChannels.sol
@@ -10,6 +10,9 @@ contract RaidenMicroTransferChannels {
      *  Data structures
      */
 
+    // The only role of the owner_address is to add or remove trusted contracts
+    address public owner_address;
+
     // Number of blocks to wait from an uncooperativeClose initiated by the sender
     // in order to give the receiver a chance to respond with a balance proof
     // in case the sender cheats. After the challenge period, the sender can settle
@@ -57,6 +60,11 @@ contract RaidenMicroTransferChannels {
      * Modifiers
      */
 
+    modifier isOwner() {
+        require(msg.sender == owner_address);
+        _;
+    }
+
     modifier isTrustedContract() {
         require(trusted_contracts[msg.sender]);
         _;
@@ -85,6 +93,9 @@ contract RaidenMicroTransferChannels {
         address indexed _receiver,
         uint32 indexed _open_block_number,
         uint192 _balance);
+    event TrustedContract(
+        address indexed _trusted_contract_address,
+        bool _trusted_status);
 
 
     /*
@@ -114,11 +125,8 @@ contract RaidenMicroTransferChannels {
         require(token.totalSupply() > 0);
 
         challenge_period = _challenge_period;
-        for (uint256 i = 0; i < _trusted_contracts.length; i++) {
-            require(_trusted_contracts[i] != 0x0);
-            require(addressHasCode(_trusted_contracts[i]));
-            trusted_contracts[_trusted_contracts[i]] = true;
-        }
+        owner_address = msg.sender;
+        addTrustedContracts(_trusted_contracts);
     }
 
     /*
@@ -163,6 +171,29 @@ contract RaidenMicroTransferChannels {
                 open_block_number,
                 deposit
             );
+        }
+    }
+
+    /// @notice Function for adding trusted contracts. Can only be called by owner_address.
+    /// @param _trusted_contracts Array of contract addresses that can be trusted to
+    /// open and top up channels on behalf of a sender.
+    function addTrustedContracts(address[] _trusted_contracts) isOwner public {
+        for (uint256 i = 0; i < _trusted_contracts.length; i++) {
+            if (addressHasCode(_trusted_contracts[i])) {
+                trusted_contracts[_trusted_contracts[i]] = true;
+                TrustedContract(_trusted_contracts[i], true);
+            }
+        }
+    }
+
+    /// @notice Function for removing trusted contracts. Can only be called by owner_address.
+    /// @param _trusted_contracts Array of contract addresses to be removed from the trusted_contracts mapping.
+    function removeTrustedContracts(address[] _trusted_contracts) isOwner public {
+        for (uint256 i = 0; i < _trusted_contracts.length; i++) {
+            if (trusted_contracts[_trusted_contracts[i]]) {
+                trusted_contracts[_trusted_contracts[i]] = false;
+                TrustedContract(_trusted_contracts[i], false);
+            }
         }
     }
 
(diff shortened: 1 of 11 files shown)


contracts/contracts/RaidenMicroTransferChannels.sol
```diff
@@ -86,7 +87,12 @@ contract RaidenMicroTransferChannels {
     /// @param _challenge_period A fixed number of blocks representing the challenge period.
     /// We enforce a minimum of 500 blocks waiting period.
     /// after a sender requests the closing of the channel without the receiver's signature.
-    function RaidenMicroTransferChannels(address _token_address, uint32 _challenge_period) public {
+    function RaidenMicroTransferChannels(
```
LefterisJP: Also add a docstring entry for `_trusted_contracts`

contracts/contracts/RaidenMicroTransferChannels.sol
```diff
@@ -140,11 +159,31 @@ contract RaidenMicroTransferChannels {
...
 
-        // transferFrom deposit from sender to contract
-        // ! needs prior approval from user
+        // transferFrom deposit from msg.sender to contract
+        // ! needs prior approval from msg.sender
+        require(token.transferFrom(msg.sender, address(this), _deposit));
+    }
+
+    /// @notice Function that allows a trusted contract to create a new channel between `_sender_address` and `_receiver_address` and transfers
+    /// the `_deposit` token deposit to this contract, compatibility with ERC20 tokens.
```
LefterisJP:
```diff
-token deposit to this contract, compatibility with ERC20 tokens.
+token deposit to this contract. Compatibility with ERC20 tokens.
```

Also add a mention that this function has to be called by a trusted contract.

contracts/contracts/RaidenMicroTransferChannels.sol
```diff
@@ -140,11 +159,31 @@ contract RaidenMicroTransferChannels {
...
+    /// @param _sender_address The sender's address in behalf of whom the delegate sends tokens.
+    /// @param _receiver_address The address that receives tokens.
+    /// @param _deposit The amount of tokens that the sender escrows.
+    function createChannelDelegate(
+        address _sender_address,
+        address _receiver_address,
+        uint192 _deposit)
+        external
+    {
+        require(trusted_contracts[msg.sender]);
```
LefterisJP: Suggestion: Turn this into a modifier. i.e.: `callerIsTrustedContract` and use that instead here and in the other functions where you have this require
```
require(trusted_contracts[msg.sender]);
```

contracts/contracts/RaidenMicroTransferChannels.sol
```diff
@@ -172,6 +211,34 @@ contract RaidenMicroTransferChannels {
...
+
+        updateInternalBalanceStructs(
+            _sender_address,
+            _receiver_address,
+            _open_block_number,
+            _added_deposit
+        );
+
+        // transferFrom deposit from msg.sender to contract
+        // ! needs prior approval from user
```
LefterisJP:
```diff
-! needs prior approval from user
+! needs prior approval from the trusted contract 
```

contracts/contracts/RaidenMicroTransferChannels.sol
```diff
@@ -97,6 +103,11 @@ contract RaidenMicroTransferChannels {
         require(token.totalSupply() > 0);
 
         challenge_period = _challenge_period;
+        for (uint256 i = 0; i < _trusted_contracts.length; i++) {
```
LefterisJP: With a large enough `_trusted_contracts` array this loop can botch the contstructor and make it run out of gas. Is that okay? What is the average expected size of this array? How many trusted contracts will we be setting for a microraiden channel manager?

LefterisJP:
As an addendum to this comment of my review:

> With a large enough `_trusted_contracts` array this loop can botch the contstructor and make it run out of gas. Is that okay? What is the average expected size of this array? How many trusted contracts will we be setting for a microraiden channel manager?

I would like to explore the possibility of adding the ability to modify the trusted contracts array. The same "owner" key as the one deploying should be the only one able to do it. 

This way if somehow a contract is botched/compromised it can be removed and new delegates can be added without redeploying the entire channel manager contract.

contracts/contracts/RaidenMicroTransferChannels.sol
loredanacirstea: Current plan is to have 1 trusted contract - ex. market maker through which a user can open a channel by sending ETH instead of the channel token.
Thought about the out of gas possibility, but at this point, even the trusted contracts are something that we control and develop. I see no reason to have more than 1-2 trusted contracts.
Should be discussed more.

loredanacirstea:
@LefterisJP 
> I would like to explore the possibility of adding the ability to modify the trusted contracts array. The same "owner" key as the one deploying should be the only one able to do it.

I agree that this should be discussed. I did not implement this because we decided on removing `owner` for the last release. 
However, we can also think about trusted contracts as 3rd party contracts that we may decide to partner with. In this case, it makes even more sense to have a separate `setTrustedContract` & `removeTrustedContract`.

contracts/contracts/RaidenMicroTransferChannels.sol
```diff
@@ -223,9 +235,8 @@ contract RaidenMicroTransferChannels {
         uint32 _open_block_number,
         uint192 _added_deposit)
         external
+        isTrustedContract
```
LefterisJP: this is a style thing and I don't think we got it specified anywhere but I believe it's easier to read if the visibility specifier comes last and before it we got the modifiers.

loredanacirstea: This was the style I used for the auction contract: visibility, then modifiers.
But it makes sense to make the visibility specifier more.. visible.

LefterisJP:
Left an extra style comment after approving.
Also we would need to discuss how trusted contracts would work. I just can't see it as not being modifiable after deploying. As you also mentioned, perhaps there is going to be a partnership with someone offering to be a delegate. We would need to add them and redeploying would not be an option.

loredanacirstea:
@LefterisJP , refactored the trusted contracts logic. Also added an event to keep track of them.

Issue #248, opened 2017-12-04
Allow receiver to withdraw tokens without closing channel


andrevmatos:
The idea is that, when needed, the receiver could go on-chain and provide latest (or some) balance proof received from the sender, and withdraw its due tokens, without closing the channel and burdening the sender on opening a new channel if they want to continue transactioning with the receiver.

One possible implementation idea is to have a mapping of channel => withdrawn amount, which would not be initialized on channel open (to avoid increasing sender's gas usage), but would be assumed as 0. On withdraw, as on close, the balance proof would be required and validated, and `balance_proof - (already_withdrawn or 0)` tokens are transfered to receiver's account (if `> 0`), and `already_withdrawn = balance_proof` is set. An event could be emitted to notify of this withdraw tx. Exactly same logic goes for channel close, but added of closing logic (release mappings allocation, emit respective events, etc).
This would allow receiver (specially those dealing with large amount of tokens/long living channels) to be liquid, without burdening senders/clients on opening channels again, allowing for very long channel's life-span.

A python script (like [close_all_channels.py](https://github.com/raiden-network/microraiden/blob/9573de930c1716a97a42ec1d470728f7ab9727ec/microraiden/microraiden/close_all_channels.py)) could be used to demonstrate usage of this feature. Nothing changes client-side.

PR #324, opened 2018-01-02
Allow receiver to withdraw tokens without closing channel

loredanacirstea:
closes https://github.com/raiden-network/microraiden/issues/248

- add `withdraw` function 
  - estimated gas cost: `72525` first withdrawal, `42589` subsequent withdrawals
- add `ChannelWithdraw` event
- `getChannelInfo` now returns
```
return (
            key,
            channels[key].deposit,
            closing_requests[key].settle_block_number,
            closing_requests[key].closing_balance,
            withdrawn_balances[key]
        );
```
- added `_receiver_tokens` to `ChannelSettled`
```
event ChannelSettled(
        address indexed _sender,
        address indexed _receiver,
        uint32 indexed _open_block_number,
        uint192 _balance,
        uint192 _receiver_tokens);
```
- tests

### DIFF
--- a/contracts/contracts/RaidenMicroTransferChannels.sol
+++ b/contracts/contracts/RaidenMicroTransferChannels.sol
@@ -291,7 +291,9 @@ contract RaidenMicroTransferChannels {
     /// @notice Allows channel receiver to withdraw tokens.
     /// @param _open_block_number The block number at which a channel between the
     /// sender and receiver was created.
-    /// @param _balance The amount of tokens owed by the sender to the receiver.
+    /// @param _balance Partial or total amount of tokens owed by the sender to the receiver.
+    /// Has to be smaller or equal to the channel deposit. Has to match the balance value from
+    /// `_balance_msg_sig` - the balance message signed by the sender.
     /// Has to be smaller or equal to the channel deposit.
     /// @param _balance_msg_sig The balance message signed by the sender.
     function withdraw(
--- a/contracts/contracts/RaidenMicroTransferChannels.sol
+++ b/contracts/contracts/RaidenMicroTransferChannels.sol
@@ -285,6 +285,7 @@ contract RaidenMicroTransferChannels {
         // ! needs prior approval from the trusted contract
         // Do transfer after any state change
         require(token.transferFrom(msg.sender, address(this), _added_deposit));
+    }
 
     /// @notice Allows channel receiver to withdraw tokens.
     /// @param _open_block_number The block number at which a channel between the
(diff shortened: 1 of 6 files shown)


loredanacirstea:
Current implementation is vulnerable to replay attacks - someone can use the same balance proof multiple times.
Initial proposal by @andrevmatos from #248 is being implemented.

contracts/contracts/RaidenMicroTransferChannels.sol
```diff
@@ -568,7 +613,7 @@ contract RaidenMicroTransferChannels {
         delete closing_requests[key];
 
         // Send _balance to the receiver, as it is always <= deposit
-        require(token.transfer(_receiver_address, _balance));
+        require(token.transfer(_receiver_address, _balance - withdrawn_balances[key]));
```
andrevmatos: Maybe (if not increase gas cost) it'd be nice to also add `_balance - withdrawn_balances[key]` to the `ChannelSettled` event, to inform the actual final transfer value besides the closing/settlement balance?

loredanacirstea: When testing with populus, there is an additional `301` gas cost for cooperativeClose. So, not that much. 
Adding this would actually give enough information for getting the state of the channel when settling, so I'm adding it to the event.

contracts/contracts/RaidenMicroTransferChannels.sol
```diff
@@ -281,6 +288,44 @@ contract RaidenMicroTransferChannels {
         require(token.transferFrom(msg.sender, address(this), _added_deposit));
     }
 
+    /// @notice Allows channel receiver to withdraw tokens.
+    /// @param _open_block_number The block number at which a channel between the
+    /// sender and receiver was created.
+    /// @param _balance The amount of tokens owed by the sender to the receiver.
+    /// Has to be smaller or equal to the channel deposit.
+    /// @param _balance_msg_sig The balance message signed by the sender.
+    function withdraw(
```
LefterisJP: Why is the receiver forced to withdraw the entire owed balance to him here? Why not provide the option to just withdraw part of it?

loredanacirstea: The receiver just has to provide a balance proof signed by the sender. 
In theory, it can be any balance proof - even an older one. In practice, the current microraiden server only stores the last balance proof. 

So, the comment might be misleading. I can change to
```
/// @param _balance Partial or total amount of tokens owed by the sender to the receiver.
/// Has to be smaller or equal to the channel deposit. Has to match the balance value
/// from `_balance_msg_sig` - the balance message signed by the sender.   
```

LefterisJP: yep that sounds better. The microraiden server implementation may change at some point and allow for partial withdrawal without any contract changes

chrouzz:
what is the status of this?
Did you finally implemented this feature?
Is it safe?

loredanacirstea:
@chrouzz , you are commenting on a PR that was adding the `withdraw` functionality. You can see that it was merged. So, yes, it was implemented.

Issue #203, opened 2017-11-28
Fix requirements.txt for contracts


loredanacirstea:
Add missing packages, versions
Move populus from main requirements.txt to the contract related one.

PR #327, opened 2018-01-04
Populus compat2

loredanacirstea:
closes https://github.com/raiden-network/microraiden/issues/203

- update to populus 2.2.0, web3 v4.0.0-beta.4 & others
- use hardcoded versions for some packaged due to dependency issues related to `cytoolz`
- use checksummed addresses instead of normalized
- use eth_utils.is_same_address for comparing addresses


contracts/README.md
```diff
@@ -40,14 +40,14 @@ The Smart Contracts can be installed separately from the other components of the
 populus compile
 
 # tests
-pytest
-pytest -p no:warnings -s
-pytest tests/test_uraiden.py -p no:warnings -s
+py.test
```
LefterisJP: `pytest` also works fine

loredanacirstea: ah, you are right. I was confused because `pytest` would use my global installation and `py.test` the local one from the virtualenv. But `py.test` is actually being deprecated in favor of `pytest`...

Issue #276, opened 2017-12-12
Contracts - upgrade to solc 0.4.19


loredanacirstea:
Includes upgrading populus version.

loredanacirstea:
Depends on populus being upgraded to `>=2.2.0` in https://github.com/raiden-network/microraiden/pull/311 .

loredanacirstea:
Not needed. Contracts can now be compiled with 0.4.19 (dependencies issues fixed), but the minimum version can stay at 0.4.17

PR #333, opened 2018-01-05
Contracts - upgrade solidity compiler version to ^0.4.19

loredanacirstea:
closes https://github.com/raiden-network/microraiden/issues/276

### DIFF
--- a/contracts/contracts/RaidenMicroTransferChannels.sol
+++ b/contracts/contracts/RaidenMicroTransferChannels.sol
@@ -1,4 +1,4 @@
-pragma solidity ^0.4.17;
+pragma solidity ^0.4.19;
 
 import './Token.sol';
 import './lib/ECVerify.sol';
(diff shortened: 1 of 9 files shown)


contracts/contracts/RaidenMicroTransferChannels.sol
```diff
@@ -1,4 +1,4 @@
-pragma solidity ^0.4.17;
+pragma solidity ^0.4.19;
```
LefterisJP: are you using some 0.4.18/0.4.19 only features? If not then I guess this can stay as is.

loredanacirstea: No, `0.4.17` should work too. I usually update this too keep up with https://etherscan.io/solcbuginfo . There is a fixed bug in 0.4.18 but it is of very low severity, though.

LefterisJP: These info appear on the verified contract, depending on the version you compiled the contract with and not the minimum required version for compiling shown in the source code.

loredanacirstea:
Closing this as it is not needed.

PR #334, opened 2018-01-05
Bump version to 0.2.0

loredanacirstea:
Feature branch

TODO
- update main net address
- update contract version & addresses in uRaiden config file


LefterisJP:
Looks fine, but I would advise this to be the last thing to be merged before tagging the release, so that it coincides with the tag.

Issue #303, opened 2017-12-18
depend on eth-tester and not ethereum-tester


gleim:
ethereum-tester is an out of date dependency and breaks new builds pulling in microraiden

pcppcp:
We've now freezed `ethereum-tester` to version `0.4.0`. The reason is an incompatible `web3` which in new version removed some of the function we use in raiden and we need to replace them.

pcppcp:
Fixed by PR #327

PR #336, opened 2018-01-08
Update ethereum-tester to eth-tester==0.1.0b5

loredanacirstea:
fixes https://github.com/raiden-network/microraiden/issues/303

This was solved for `requirements.txt`, but not for `requirements-dev.txt` in  https://github.com/raiden-network/microraiden/pull/327

Version is strictly specified due to version dependency issues related to the `cytoolz` package.

### DIFF
--- a/microraiden/requirements-dev.txt
+++ b/microraiden/requirements-dev.txt
@@ -1,5 +1,5 @@
 -r requirements.txt
 pytest>=2.7.2,!=3.3.*
-ethereum-tester
+eth-tester==0.1.0b5
 mock
 requests_mock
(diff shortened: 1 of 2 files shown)


microraiden/requirements-dev.txt
```diff
@@ -1,5 +1,5 @@
 -r requirements.txt
 pytest>=2.7.2,!=3.3.*
-ethereum-tester
+eth-tester==0.1.0b5
```
pcppcp: shouldn't we also remove the tester from `requirements.txt`?

loredanacirstea: Yes. Instead of replacing `ethereum-tester` from requirements-dev.txt I added `eth-tester` in requirements.txt by mistake in a previous PR and forgot about this.
Fixed now.

PR #339, opened 2018-01-08
Enable travis test of js/ts library

andrevmatos:


andrevmatos:
Agree with the fact that it's annoying to have it asking to commit the changes (even if it usually is wrong, because different origin remote or so). It was the only place where the documentation was available (that's why `.md` was choosen), but as soon as it's on readthedocs, I'll remove and `.gitignore` it here.

PR #344, opened 2018-01-10
Update config.py

nickvonh:
I have updated the channel manager address, as it currently causes this error in the echo_server.py example.
```
File "/root/micro-raiden/micro-raiden/env/lib/python3.5/site-packages/web3/utils/validation.py", line 117, in validate_address
    raise InvalidAddress("Address has an invalid EIP checksum", value)
web3.exceptions.InvalidAddress: ('Address has an invalid EIP checksum', '0x161a0d7726eb8b86eb587d8bd483be1ce87b0609')
```

got address from https://ropsten.etherscan.io/address/0x161a0d7726eb8b86eb587d8bd483be1ce87b0609

### DIFF
--- a/microraiden/microraiden/config.py
+++ b/microraiden/microraiden/config.py
@@ -21,7 +21,7 @@ NETWORK_NAMES = {
 
 # address of the default channel manager contract. You can change this using commandline
 # option --channel-manager-address when running the proxy
-CHANNEL_MANAGER_ADDRESS = '0x161a0d7726eb8b86eb587d8bd483be1ce87b0609'  # ropsten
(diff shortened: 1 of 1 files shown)


pcppcp:
Hi @nickvonh 
thanks for your contribution! After updating the `web3` python library we've decided to use checksummed addresses and this caused some problems, like this one you've discovered.
Our goal is to accept non-checksummed addresses on any entrypoint (REST interface, public API methods..) to make the library easy to use.
I already have a change that fixes `echo_server` startup prepared and will merge it soon. There's no need to change the `config.py` now as we'll update contract version sometime this week.

PR #347, opened 2018-01-11
Tests can be run on Ropsten

devium:
This PR mainly introduces some compatibility changes and command line options to the test environment to properly run the tests on a real testnet (tested with Ropsten).

To run on the testnet:
* Start a testnet node on `localhost:8545`
* Pass `--no-tester` to `pytest`
* Pass `--faucet-private-key <key>` to `pytest` where `<key>` is either a hex-encoded private key or a path to a JSON file. This account will be used to fund transactions with ETH and testnet RDN tokens. Total transaction cost should be below 0.5 ETH. Remaining ETH and RDN will be returned to this address.
* If the faucet private key is a JSON file, make sure to pass `--faucet-password-path <path>` for a file containing the private key password.
* Optionally: Use a different tester private key generator seed with `--private-key-seed <int>` if you're scared of loss of funds or want to avoid collisions with simultaneously running tests.

This PR also includes several bug fixes encountered along the way, most of them having to do with the switch from `RPCProvider` to `HTTPProvider` and consistent use of checksummed addresses.

WIP: most tests still fail on testnet.

### DIFF
--- a/microraiden/microraiden/channel_manager/manager.py
+++ b/microraiden/microraiden/channel_manager/manager.py
@@ -7,7 +7,8 @@ import logging
 import os
 from eth_utils import (
     decode_hex,
-    is_same_address
+    is_same_address,
+    is_checksum_address
 )
 from ethereum.exceptions import InsufficientBalance
 from web3 import Web3
@@ -126,6 +127,7 @@ class ChannelManager(gevent.Greenlet):
 
     def event_channel_opened(self, sender, open_block_number, deposit):
         """Notify the channel manager of a new confirmed channel opening."""
+        assert is_checksum_address(sender)
         if (sender, open_block_number) in self.channels:
             return  # ignore event if already provessed
         c = Channel(self.state.receiver, sender, deposit, open_block_number)
@@ -136,6 +138,7 @@ class ChannelManager(gevent.Greenlet):
 
     def unconfirmed_event_channel_opened(self, sender, open_block_number, deposit):
         """Notify the channel manager of a new channel opening that has not been confirmed yet."""
+        assert is_checksum_address(sender)
         event_already_processed = (sender, open_block_number) in self.unconfirmed_channels
         channel_already_confirmed = (sender, open_block_number) in self.channels
         if event_already_processed or channel_already_confirmed:
@@ -148,8 +151,9 @@ class ChannelManager(gevent.Greenlet):
                       sender, open_block_number)
 
     def event_channel_close_requested(self, sender, open_block_number, balance, settle_timeout):
-        assert settle_timeout >= 0
         """Notify the channel manager that a the closing of a channel has been requested."""
+        assert is_checksum_address(sender)
+        assert settle_timeout >= 0
         if (sender, open_block_number) not in self.channels:
             self.log.warning(
                 'attempt to close a non existing channel (sender %ss, block_number %ss)',
@@ -175,6 +179,7 @@ class ChannelManager(gevent.Greenlet):
 
     def event_channel_settled(self, sender, open_block_number):
         """Notify the channel manager that a channel has been settled."""
+        assert is_checksum_address(sender)
         self.log.info('Forgetting settled channel (sender %s, block number %s)',
                       sender, open_block_number)
         self.state.del_channel(sender, open_block_number)
@@ -183,6 +188,7 @@ class ChannelManager(gevent.Greenlet):
             self, sender, open_block_number, txhash, added_deposit
     ):
         """Notify the channel manager of a topup with not enough confirmations yet."""
+        assert is_checksum_address(sender)
         if (sender, open_block_number) not in self.channels:
             assert (sender, open_block_number) in self.unconfirmed_channels
             self.log.info('Ignoring unconfirmed topup of unconfirmed channel '
@@ -198,6 +204,7 @@ class ChannelManager(gevent.Greenlet):
 
     def event_channel_topup(self, sender, open_block_number, txhash, added_deposit):
         """Notify the channel manager that the deposit of a channel has been topped up."""
+        assert is_checksum_address(sender)
         self.log.info(
             'Registering deposit top up (sender %s, block number %s, added deposit %s)',
             sender, open_block_number, added_deposit
@@ -219,6 +226,7 @@ class ChannelManager(gevent.Greenlet):
 
     def close_channel(self, sender, open_block_number):
         """Close and settle a channel."""
+        assert is_checksum_address(sender)
         if not (sender, open_block_number) in self.channels:
             self.log.warning(
                 "attempt to close a non-registered channel (sender=%s open_block=%s" %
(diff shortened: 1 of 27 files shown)


microraiden/microraiden/examples/echo_client.py
```diff
@@ -39,7 +37,7 @@ def run(
         key_path: str,
```
loredanacirstea: Can we change `--key-path` to `--private-key`? Reason: consistency with the other examples.

devium:
I managed to successfully run all tests on Kovan with this PR. Tests that require waiting for the settlement period to end are excluded. Other than that, everything seems to work fine.

microraiden/microraiden/client/client.py
```diff
@@ -42,7 +42,11 @@ def __init__(
         # Create web3 context if none is provided, either by using the proxies' context or creating
         # a new one.
         if not web3:
-            web3 = Web3(HTTPProvider())
+            web3 = Web3(HTTPProvider('http://localhost:8545'))
```
loredanacirstea: How about using `config.WEB3_PROVIDER_DEFAULT` here?

pcppcp: Endpoint URL should be probably configurable.

microraiden/microraiden/channel_manager/manager.py
```diff
@@ -348,6 +359,7 @@ def verify_balance_proof(self, sender, open_block_number, balance, signature):
 
     def register_payment(self, sender, open_block_number, balance, signature):
         """Register a payment."""
+        assert is_checksum_address(sender)
```
pcppcp: I prefer converting address to checksummed one in public api. They should be liberal.

microraiden/microraiden/client/client.py
```diff
@@ -171,15 +175,16 @@ def open_channel(self, receiver_address: str, deposit: int):
...
             argument_filters=filters
         )
 
         if event:
             log.debug('Event received. Channel created in block {}.'.format(event['blockNumber']))
             channel = Channel(
                 self.context,
-                event['args']['_sender_address'],
-                event['args']['_receiver_address'],
+                self.context.address,
```
pcppcp: Maybe add assert to test if event addr is equal to stored one.

microraiden/microraiden/client/client.py
devium: It is, by passing a different `Web3` object. Although @loredanacirstea's suggestion of using a config default might be more elegant.

microraiden/microraiden/config.py
```diff
@@ -21,7 +21,11 @@
 
 # address of the default channel manager contract. You can change this using commandline
 # option --channel-manager-address when running the proxy
-CHANNEL_MANAGER_ADDRESS = '0x161a0d7726eb8b86eb587d8bd483be1ce87b0609'  # ropsten
+CHANNEL_MANAGER_ADDRESS = {
```
pcppcp: We'll need to store block number in which the contract has been deployed. But this can be done in a separate PR later.
Wouldn't it be better to use simple ints as dict keys instead of string?

microraiden/microraiden/examples/echo_client.py
```diff
@@ -10,45 +10,43 @@
...
-from microraiden.config import CHANNEL_MANAGER_ADDRESS
-
 
 @click.command()
 @click.option(
-    '--key-path',
+    '--private-key',
     required=True,
-    help='Path to private key file.',
+    help='Path to private key file or a hex-encoded private key.',
```
pcppcp: It's not the smartest idea to expose privkey to anyone who can run `ps -aux` on your system :)

microraiden/microraiden/utils/contract.py
```diff
@@ -23,7 +23,7 @@ def create_signed_transaction(
         data=b'',
         nonce_offset: int = 0,
         gas_price: int = GAS_PRICE,
-        gas_limit: int = GAS_LIMIT
+        gas_limit: int = 21000
```
pcppcp: constant is much better here IMO

microraiden/microraiden/channel_manager/manager.py
devium: I'm not entirely sure about how to handle this either. I think ethereum's overdue for making checksummed addresses mandatory, so I tried to enforce them in our internal, transaction-making classes here.

Simply converting all addresses to checksummed addresses defeats the purpose of the whole notion.

microraiden/microraiden/config.py
devium: It's just for convenience, since `web3.version.network` returns strings. Since we're not using anything else for determining the network, I chose this shortcut of directly using strings.

microraiden/microraiden/examples/echo_client.py
devium: Still, developers should be able to for quick testing which is probably the case 99% of the time this example runs. I doubt people use their real mainnet wallet for these kind of tests. Even then, they'd have to get access to their hex private key.

microraiden/microraiden/config.py
```diff
@@ -3,8 +3,8 @@
 from eth_utils import denoms
 
 API_PATH = "/api/1"
-GAS_LIMIT = 250000
-GAS_PRICE = 20 * denoms.gwei
+GAS_LIMIT = 130000
+GAS_PRICE = 50 * denoms.gwei
```
loredanacirstea: What was the reason for increasing the gas price? Running the Ropsten tests faster?
I see `GAS_PRICE` is used for `create_signed_contract_transaction` when closing all channels. This should be something configurable by the user through a cli option in this case.
Do we use the gas price for anything else production-related?

microraiden/microraiden/utils/contract.py
devium: Will add.

microraiden/microraiden/channel_manager/manager.py
pcppcp: Sure, internals it should be checksummed addr (as is the case in `blockchain.py` for example)
On the other hand, channel manager _may_ be used by some courageous fool to handle payments with a protocol other than HTTP, thus I'd say we can consider it a public API.
Having `is_checksum_address` now isn't a blocker of course, but we should be prepared to support other bizzare use cases in the future.

microraiden/microraiden/config.py
devium: Yes, faster inclusion on Ropsten. Should probably decrease that again or use different prices for testnets and mainnet again.

I'd separate a new CLI option into a different PR though (+issue).

microraiden/microraiden/client/client.py
devium: Done.

microraiden/microraiden/config.py
pcppcp: we should also discuss having gas prices configurable per-network, as is the case with block number

loredanacirstea: agreed, a separate PR would be better. I created an issue https://github.com/raiden-network/microraiden/issues/354

microraiden/microraiden/client/client.py
devium: Added.

PR #356, opened 2018-01-12
CHANGELOG.md

devium:
WIP. I will go through the PRs and add them to the list. If you feel like I missed something, let me know.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,25 @@
 
 ## Development
 * Added this changelog. #356
+* Updated documentation of a µRaiden quickstart in README.md. #342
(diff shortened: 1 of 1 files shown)


devium:
So I added changelog updates for every merged PR. We could probably prune that a bit.

Also, open points of discussion:
* Listing PRs in ascending order might make more sense.
* I would recommend requiring all future PRs to update the `CHANGELOG.md` file to include said PR with a short summary, including its PR number.
* We might want to exclude bug fixes / refactorings / reversals that occurred and have been fixed in between releases, as those might be confusing to users only working on release versions.

loredanacirstea:
I agree with @pcppcp about updating the changelog with significant changes only.

PR #358, opened 2018-01-12
Automatically skip some tests

pcppcp:
This currently covers tests that depend on xorg or testnet environment.
The big deposit test is skipped until we have separate test configurations for different networks.

### DIFF
--- a/microraiden/microraiden/test/test_examples.py
+++ b/microraiden/microraiden/test/test_examples.py
@@ -2,6 +2,7 @@ import pytest  # noqa: F401
 from _pytest.monkeypatch import MonkeyPatch
 from flask import jsonify
 from web3 import Web3
+import os
 
 from microraiden import Session
 from microraiden.channel_manager import ChannelManager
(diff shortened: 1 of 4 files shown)


microraiden/microraiden/test/test_examples.py
```diff
@@ -13,7 +14,10 @@
 import microraiden.client.session
 
 
-@pytest.mark.needs_xorg
+@pytest.mark.skipif(
+    ('TEST_SKIP_XORG' in os.environ) or ('DISPLAY' not in os.environ),
```
loredanacirstea: When would you have `os.environ['DISPLAY']` set?

pcppcp: It tests for 1) existence of the tkinter library that's required to display price widget and 2) $DISPLAY that's set for a shells spawned from X
Travis has `tkinter` installed but doesn't export `$DISPLAY`...
Now that I am thinking of it, I'll check the variable in `conftest.py`, better to have it in one place.

Issue #363, opened 2018-01-12
compile webui automatically if doing `pip install -e .` or `python setup.py build`


pcppcp:

pcppcp:
Fixed #366

PR #366, opened 2018-01-15
Add npm compilation to `setup.py`

pcppcp:
Also bumped microraiden version.

### DIFF
--- a/microraiden/microraiden/config.py
+++ b/microraiden/microraiden/config.py
@@ -66,7 +66,7 @@ with open(os.path.join(MICRORAIDEN_DIR, 'microraiden', CONTRACTS_ABI_JSON)) as m
 
 # required version of the deployed contract at CHANNEL_MANAGER_ADDRESS.
 # Proxy will refuse to start if the versions do not match.
-MICRORAIDEN_VERSION = "0.1.0"
+MICRORAIDEN_VERSION = "0.1.1"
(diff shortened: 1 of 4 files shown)


pcppcp:
Kinda fixes #363 . Develop installs still require manual compilation of the npm package (`py setup.py compile_webui`).

microraiden/microraiden/config.py
```diff
@@ -66,7 +66,7 @@
 
 # required version of the deployed contract at CHANNEL_MANAGER_ADDRESS.
 # Proxy will refuse to start if the versions do not match.
-MICRORAIDEN_VERSION = "0.1.0"
+MICRORAIDEN_VERSION = "0.1.1"
```
loredanacirstea: The next release version will be `0.2.0`. Why the version change now?

pcppcp: I've accidentally pushed package without javascript library to the pypi and to fix it I had to bump maintenance version to be able to re-upload the package. Now I realize that string from `config.py` has to be equal to contract version.

pcppcp: We should probably compare only major+minor version of the contract and package, otherwise it'll be painful to do any maintenance updates.

loredanacirstea: agreed.

PR #371, opened 2018-01-16
Config gas price

### LINKED ISSUE #354: Support configurable gas price
As discussed in https://github.com/raiden-network/microraiden/pull/347#discussion_r161204245

we should consider:
- having a cli option for providing the gas price
- having default gas prices configurable per-network

pcppcp:
This PR makes it possible to configure network-related variables.
These include

- channel manager address
- start sync block
- gas price
- gas limit
- pot_gas_limit

I've moved most of the stuff from `config.py` to `constants.py`, as these are generally things that don't need frequent changes. `config.py` now includes network configuration and exports a global: `NETWORK_CFG` 

Also fixes #354

### DIFF
--- a/microraiden/microraiden/config.py
+++ b/microraiden/microraiden/config.py
@@ -24,13 +24,17 @@ NetworkConfig = partial(
 # network-specific configuration
 NETWORK_CONFIG_DEFAULTS = {
     1: NetworkConfig(
-        channel_manager_address='0x0',
-        start_sync_block=0,
-        gas_price=50 * denoms.gwei
+        channel_manager_address='0x4d6e0922e6b703f0fdf92745343a9b83eb656402',
+        start_sync_block=4651176,
+        gas_price=20 * denoms.gwei
     ),
     3: NetworkConfig(
         channel_manager_address='0x161a0d7726EB8B86EB587d8BD483be1CE87b0609',
         start_sync_block=2400640
+    ),
+    4: NetworkConfig(
+        channel_manager_address='0x568a0d52a173f584d4a286a22b2a876911079e15',
+        start_sync_block=1338285
     ),
     42: NetworkConfig(
         channel_manager_address='0xB9721dF0e024114e7B25F2cF503d8CBE3D52b400',
--- a/microraiden/microraiden/config.py
+++ b/microraiden/microraiden/config.py
@@ -1,79 +1,61 @@
-import json
-import os
 from eth_utils import denoms
-from collections import namedtuple
+from collections import namedtuple, OrderedDict
+from functools import partial
 
-API_PATH = "/api/1"
-GAS_LIMIT = 130000
 
-# Plain old transaction, for lack of a better term.
-POT_GAS_LIMIT = 21000
+# these are default values for network config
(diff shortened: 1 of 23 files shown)


microraiden/microraiden/constants.py
```diff
@@ -0,0 +1,64 @@
...
+TOKEN_ABI_NAME = 'CustomToken'
+# compiled contracts path
+CONTRACTS_ABI_JSON = 'data/contracts.json'
+
+with open(os.path.join(MICRORAIDEN_DIR, 'microraiden', CONTRACTS_ABI_JSON)) as metadata_file:
+    CONTRACT_METADATA = json.load(metadata_file)
+
+# required version of the deployed contract at CHANNEL_MANAGER_ADDRESS.
+# Proxy will refuse to start if the versions do not match.
+MICRORAIDEN_VERSION = "0.1.0"
```
loredanacirstea: Current master has `MICRORAIDEN_VERSION = "0.1.1"`

microraiden/microraiden/constants.py
```diff
@@ -0,0 +1,64 @@
...
+CHANNEL_MANAGER_CONTRACT_VERSION = MICRORAIDEN_VERSION
```
loredanacirstea: Maybe this should be set separately: `CHANNEL_MANAGER_CONTRACT_VERSION = "0.1.0"`

microraiden/microraiden/constants.py
```diff
@@ -0,0 +1,64 @@
...
+# they should stay the same until we decide otherwise
+assert MICRORAIDEN_VERSION == CHANNEL_MANAGER_CONTRACT_VERSION
```
loredanacirstea: Does this still makes sense here? If yes, maybe `check_contract_version` can be made more abstract and used here.

microraiden/microraiden/config.py
```diff
@@ -1,80 +1,63 @@
...
+    ),
+    **network_config_defaults
+)
+
+# network-specific configuration
+NETWORK_CONFIG_DEFAULTS = {
+    1: NetworkConfig(
+        channel_manager_address='0x0',
+        start_sync_block=0,
+        gas_price=50 * denoms.gwei
```
loredanacirstea: 50 gwei was needed for Ropsten to run the tests faster. This is too much for main net (20 gwei default is fine)

microraiden/microraiden/config.py
```diff
@@ -1,80 +1,63 @@
...
+        network_config_defaults
+    ),
+    **network_config_defaults
+)
+
+# network-specific configuration
+NETWORK_CONFIG_DEFAULTS = {
+    1: NetworkConfig(
+        channel_manager_address='0x0',
+        start_sync_block=0,
```
loredanacirstea: I think we can already put `4651176` from the current deployed contract, in case we forget.

microraiden/microraiden/constants.py
pcppcp: ok, I'll set them separatedly. It doesn't make sense for them to be the same now, we check major+minor version anyway.

PR #388, opened 2018-01-19
Headers cleanup

loredanacirstea:
- remove deprecated `RDN-Insufficient-Funds` header
- replace with invalid proof header in the python client
- move header & cookies ValueError check after values are updated, to also catch invalid cookie values


loredanacirstea:
Looks good.

PR #391, opened 2018-01-21
examples issue fix and added missing constants values

jakjothi:


microraiden/microraiden/client/client.py
```diff
@@ -13,7 +13,7 @@
     create_signed_contract_transaction
 )
 
-import microraiden.config as config
+from microraiden.constants import CHANNEL_MANAGER_ADDRESS
```
pcppcp: Channel manager address should be imported from `constants` submodule now, after setting current network global defaults. That's because uRaiden now supports per-network configuration for addresses, gas price and so on.

See i.e. `click_helpers.py`

```python
NETWORK_CFG.set_defaults(int(web3.version.network))
channel_manager_address = to_checksum_address(
                channel_manager_address or NETWORK_CFG.CHANNEL_MANAGER_ADDRESS
)
```

Issue #392, opened 2018-01-22
not check withdraw_balances le _balance in settleChannel.


nakajo2011:
in withdraw: check withdraw_balances[key] less than _balance
https://github.com/raiden-network/microraiden/blob/01346c05ac42abb9e4534888a64e0ca82d7fe448/contracts/contracts/RaidenMicroTransferChannels.sol#L302

but, in settleChannel: not check.
https://github.com/raiden-network/microraiden/blob/01346c05ac42abb9e4534888a64e0ca82d7fe448/contracts/contracts/RaidenMicroTransferChannels.sol#L631-L641

i think need same check in settleChannel.

loredanacirstea:
@nakajo2011 , thank you for observing this! 

We have implemented this change here: https://github.com/raiden-network/microraiden/pull/394.
While it is not required (see PR description), this check is good for clarity and helps to keep track of invariants.

PR #394, opened 2018-01-22
Add underflow check for failing faster.

loredanacirstea:
Issue raised in https://github.com/raiden-network/microraiden/issues/392 (add `require(withdrawn_balances[key] < _balance);` check in `settleChannel`)
To be discussed if this is really necessary.

In case the balance that has already been withdrawn is bigger than the closing balance, `settleChannel` would still fail without this underflow check when trying to transfer tokens to the receiver:
https://github.com/raiden-network/microraiden/blob/01346c05ac42abb9e4534888a64e0ca82d7fe448/contracts/contracts/RaidenMicroTransferChannels.sol#L642

In case of underflow:
```
receiver_remaining_tokens > token_instance.call().balanceOf(uraiden_instance.address)
```
due to already withdrawn tokens -> transaction would fail anyway.

### DIFF
--- a/contracts/contracts/RaidenMicroTransferChannels.sol
+++ b/contracts/contracts/RaidenMicroTransferChannels.sol
@@ -630,6 +630,7 @@ contract RaidenMicroTransferChannels {
 
         require(channel.open_block_number > 0);
         require(_balance <= channel.deposit);
+        require(withdrawn_balances[key] <= _balance);
 
         // Remove closed channel structures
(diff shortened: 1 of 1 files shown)


loredanacirstea:
@LefterisJP , can you weigh in on this please?

LefterisJP:
I also think the contract is safe without this underflow check but I like explicitly stating the invariables

PR #398, opened 2018-01-22
Withdraw script

### LINKED ISSUE #330: Add withdraw feature, similar to close_all_channels
Contract now has a `withdraw` function for the receiver.
```
function withdraw(
        uint32 _open_block_number,
        uint192 _balance,
        bytes _balance_msg_sig)
        external
    {...}
```

As suggested in https://github.com/raiden-network/microraiden/issues/248, a script for withdrawing tokens can be added. Maybe with some parameters such as: time since last withdraw, min number of tokens to withdraw etc.

andrevmatos:
And fixes & improvements to close_all_channels.py script. Both tested in Ropsten, with live example, and working against current master. Fix #330

### DIFF
--- a/microraiden/microraiden/test/fixtures/client.py
+++ b/microraiden/microraiden/test/fixtures/client.py
@@ -48,7 +48,6 @@ def client(
             client,
             private_keys,
             channel_manager_address,
-            balance=0
         )
 
     yield client
@@ -58,5 +57,4 @@ def client(
             client,
             private_keys,
             channel_manager_address,
-            balance=0
         )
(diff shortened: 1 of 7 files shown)


microraiden/microraiden/test/fixtures/client.py
```diff
@@ -48,7 +48,6 @@ def client(
             client,
             private_keys,
             channel_manager_address,
-            balance=0
```
andrevmatos: @pcppcp withdraw was failing because of this: although a withdraw was made in the test, teardownn tried to set balance=0 on close, which made the tx fails. Is it ok to simply remove it, avoiding updating the balance in the client, and closing with current balance?

microraiden/microraiden/test/test_client.py
```diff
@@ -75,6 +75,8 @@ def test_integrity(client: Client, receiver_address):
     c.update_balance(100)
     assert not c.is_valid()
 
+    c.update_balance(0)
```
andrevmatos: Now, this invalid balance needs to be updated before proceeding with teardown.

PR #407, opened 2018-01-24
Use bumpversion for versioning

pcppcp:
Version is stored in a plaintext `microraiden/VERSION` file.
The reason is that it is easier to read - storing version in `constants.py` file makes it impossible to import `microraiden` module if dependencies are not installed.

This is more like a hotfix, later we can use something like `bumpversion` for versioning (it has git tag support and other cool features)


microraiden/microraiden/VERSION
```diff
@@ -0,0 +1 @@
+0.2.1
```
devium: Maybe we should use a dev version to avoid confusion on bug reports etc. Something like 0.2.1-DEV. Only use proper versions in release commits. Thoughts?

microraiden/microraiden/__init__.py
```diff
@@ -1,9 +1,9 @@
 # to fix requests's SSL infinite recursion bug, SSL must be
 # monkeypatched first, and only then requests can be imported
 # the bug occurs with Python3.6, requests=2.8.14 and gevent==1.2.2
-from gevent import monkey
-monkey.patch_ssl()
-import requests # noqa
+# from gevent import monkey
```
devium: Remove please if this is not needed.

microraiden/microraiden/VERSION
pcppcp: for dev version it's usual to append current HEAD checksum (like `0.2.1-a1c8defa)

I gave it a little thought and it may work like this:
1) `setup.py` needs to obtain the version. For packaging, it can check if `VERSION` file exists. In developement, the version will be output of `git describe --tag`.
2) when building the package (`py setup.py sdist`), the script can compile that `VERSION` file. It won't be git-tracked -- it'll be package specific.
3) nice thing about `git describe` is that it returns something like this:
`v0.1.4-15-g7c5925f`
where `v0.1.4` is obviously a tag
`15` is number of commits since the tag
`g7c5925f` is a current commit hash

So for 'proper' pypi releases the reported version will always be simple tag (like `v0.2.1`) and for people using git repo we'll see which commit they are on.

LefterisJP:
> This is more like a hotfix, later we can use something like bumpversion for versioning (it has git tag support and other cool features)

@pcppcp @devium Definitely recommend using `bumpversion`. We are already [doing so in raiden](https://github.com/raiden-network/raiden/blob/master/.bumpversion.cfg)

pcppcp:
@LefterisJP 
okay, I am updating versioning to use `bumpversion`. Seems to be a very useful tool so far.

microraiden/setup.py
```diff
@@ -15,14 +15,43 @@
...
+        if git_version.startswith('v'):
+            git_version = git_version[1:]
+
+        git_version = git_version.strip()
+        # if this is has commits after the tag, it's a prerelease:
+        if git_version.count('-') == 2:
+            _, _, commit = git_version.split('-')
+            if commit.startswith('g'):
+                commit = commit[1:]
+            return '{}+git.r{}'.format(VERSION, commit)
```
loredanacirstea: So, in case we have intermediary releases, the version would look like this: `0.2.0+git.rc80fc57`?
I personally like the `0.2.0-rc.2` pattern more - increasing the number makes the release order clearer.

pcppcp: No, for pypi fix releases we'll bump patch version. Releases' version will never use `+git.*` suffix.
The `git` tag is only useful i.e. if someone uses the git code directly and files a bug report - then we know which commit he is on.

Release version bumping is simple - just branch out and do `bumpversion minor && git push` . This will automatically increment desired release number (major, minor or patch) and tag, update & commit all affected files (`setup.py` and `.bumpversion.cfg`)

loredanacirstea: Sorry, when I said intermediary releases I was thinking of prereleases - between two patch versions (so, using`-rc.2` instead of `+git.rc80fc57`). I thought you also wanted to publish them. But yes, git tags are useful for debugging.

PR #409, opened 2018-01-27
enhanced documentation for web-ui example

pycal:
added clone step
corrected path to web-ui component

Signed-off-by: Cailen McQuattie


andrevmatos:
The README file is in the root of the repository. It's expected that the user knows how to clone a git repository and actually is already inside the folder, so the previous instructions are correct. Anyway, there may be other forms of fetching the code tree, like zip release packages and package managers, so the instructions must be agnostic, and therefore we want to avoid this step. Anyway, thank you very much for your contribution, and we look ahead for new ones.

Issue #414, opened 2018-02-01
deploying microraiden on Rinkeby


sbaks0820:
I'm trying to get deploy the contracts for microraiden on Rinkeby according to the instructions [in the docs](http://microraiden.readthedocs.io/en/latest/contract/index.html#opening-a-transfer-channel). But I encounter the following error executing 
```
python3 -m deploy.deploy_testnet --chain rinkeby --owner 0xe5e3a76c406321e15039dadccb1224e49776592d --supply 1000000 --token-decimals 18 --challenge-period 500 --token-name CustomToken --token-symbol FCK
```

```
Make sure rinkeby chain is running, you can connect to it and it is synced,
          or you'll get timeout
Web3 provider is RPC connection http://172.19.0.17:8745
Owner is 0xe5e3a76c406321e15039daDcCB1224e49776592D
Traceback (most recent call last):
  File "/home/suryab/projects/thesis/env/lib/python3.6/site-packages/web3/utils/formatters.py", line 66, in apply_formatters_to_dict
    yield key, formatters[key](item)
  File "cytoolz/functoolz.pyx", line 232, in cytoolz.functoolz.curry.__call__
  File "/home/suryab/projects/thesis/env/lib/python3.6/site-packages/web3/middleware/pythonic.py", line 94, in to_hexbytes
    result, len(result), num_bytes
ValueError: The value HexBytes('0xd783010800846765746887676f312e392e32856c696e75780000000000000000913a63403b7083000106cd0be9b41c9aa772cb28530d6bf732ba57c8f77088db2a0e6edcdb03e494e85b4e4964053f9a5822714fa02bf25dfecefb2325adece701') is 97 bytes, but should be 32

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/usr/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/suryab/projects/thesis/microraiden/contracts/deploy/deploy_testnet.py", line 113, in <module>
    main()
  File "/home/suryab/projects/thesis/env/lib/python3.6/site-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/home/suryab/projects/thesis/env/lib/python3.6/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
[...]

PR #416, opened 2018-02-01
Temporary fix for Rinkeby - PoA extraData bytes

loredanacirstea:
fixes https://github.com/raiden-network/microraiden/issues/414

Temporary fix for Rinkeby; PoA adds bytes to extraData, which is not yellow-paper-compliant.
https://github.com/ethereum/web3.py/issues/549


loredanacirstea:
Not merging this because it is not enough for the whole system to work. This should be fixed in web3.py

PR #421, opened 2018-02-08
Update package.json

M-Igashi:
adding latest types/lodash

on npm@5.6.0, node@v9.5.0, Ubuntu16.04,
Build error occurs in typescript by saying;
```
node_modules/@types/lodash/index.d.ts:12651:53 - error TS2344: Type 'T' does not satisfy the constraint 'object'.

12651         isWeakSet<T>(value?: any): value is WeakSet<T>;
```

with 
```sudo npm install --save @types/lodash```
the build finishes without significant errors.


andrevmatos:
Yes, reproduced it, and is caused by typescript@2.7. Plase, sign the cla so we can merge your PR.

PR #422, opened 2018-02-15
Restructure project, make requirements explicit

ezdac:
The microraiden python package was moved one level up to the
root directory. The setup.py for microraiden now resides in the
root directory.
Parsing of a requirements-file in setup.py was replaced by an
explicit list of version fixes, as needed for setuptools 'install_requires'.
Only deployment/development specific packages that are not neccessarily
needed for installation should be specified in requirements files.

Also, there where install-scripts added in the Makefile for easy
user installation.


pcppcp:
Have you tried building a package? (`python setup.py sdist`)

ezdac:
Yes, building a package is working, anything else to be done with this?

pcppcp:
Just rebase. It's a lot of changes, I'll clone your branch and check what's changed.

pcppcp:
looks good so far, but since it's a lot of changes it would be great to have thumbs up also from other members of the team

ezdac:
sure, especially the relocating of requirements is possibly an opinionated matter

pcppcp:
hm, I think having the requirements in a separate file is better, and consistent with other projects we have. It's also easier to test library compatibility with `pip -r requirements.txt`.
Was there any particular reason why did you put it to `setup.py`?

ezdac:
Mostly because a lot of sources recommend using requirement files only for freezing of deployment environments (with possible addtitional packages that aren't needed for installation, e.g. bumpversion in the dev-requirements).
Then, a naive line-by-line parsing of requirement files restricts the format of those files to exactly the form of the `install_requires` list. No usage of comments or git-source-urls or any other req-file feature are allowed in this scenario.

Getting rid of simply outsourcing of `install_requires` into the req-file frees us to use it with pipenv/piptools later on for better dependency management.

look at this post for an idea:
https://caremad.io/posts/2013/07/setup-vs-requirement/

loredanacirstea:
Smart contract-related commands also work. Looks good.

ezdac:
awesome

PR #426, opened 2018-02-19
Update docs config info

loredanacirstea:


docs/conf.py
```diff
@@ -62,17 +62,17 @@
...
 # The version info for the project you're documenting, acts as replacement for
 # |version| and |release|, also used in various other places throughout the
 # built documents.
 #
 # The short X.Y version.
-version = '0.1'
+version = '0.2'
 # The full version, including alpha/beta/rc tags.
-release = '0.1.0'
+release = '0.2.0'
```
pcppcp: can't we simply extract the version from `microraiden/VERSION` ?

loredanacirstea: Sure, but I did not want to touch this too much, knowing that @ezdac is currently working on the docs. I just wanted to get rid of the legal issues asap.

ezdac: Extraction is implemented in the `docs-develop` branch.
This hotfix won't be enough for RTD to build, since it currently is missing a `docs/requirements.txt` file. I'll try to do free some time and do a PR this evening that this can be resolved asap.

loredanacirstea: Ok, version is not that important for the scope of this PR, so I'm merging it.

PR #417, opened 2018-02-01
Autodoc feature and doc update

ezdac:
- This PR will make autodoc work and classes, methods and modules can now be referenced in the documentation as directives.
- Also, there is an automatic documentation of command-line options from `click`.
- Most of the README files were stripped down and the content transferred to the documentation.

PR #437, opened 2018-02-26
Merge new microraiden documentation version

seichris:
- Restructured documentation
- New setup tutorial
- Text & wording fixes


docs/applications/pyserver/index.rst
```diff
@@ -0,0 +1,16 @@
...
+
+
+
+.. toctree::
+   :maxdepth: 2 
+
+   installation
+   usage
+   cmdline
+
```
loredanacirstea: I see there is no `py-api.rst` - why not?

loredanacirstea: Items from`Specifications` should be moved here, after command line options - those specs are proxy-server related.
```
   installation
   usage
   cmdline
   rest-api
   http-headers
   offchain sequence
   components-overview
```

docs/applications/pyserver/usage.rst
```diff
@@ -0,0 +1,26 @@
...
+    python3 -m microraiden.examples.wikipaydia --private-key <private_key_file> --private-key-password-file <password_file> start
+
+By default, the web server listens on ``0.0.0.0:5000``. The private key
+file should be in the JSON format produced by Geth/Parity and must be
+readable and writable only by the owner to be accepted (``-rw-------``).
+A ``--private-key-password-file`` option can be specified, containing
+the password for the private key in the first line of the file. If it's
+not provided, the password will be prompted interactively. An Ethereum
+node RPC interface is expected to respond on http://localhost:8545.
+Alternatively, you can use `Infura
```
loredanacirstea: afaik, you could not use Infura. Did someone test and it worked?

docs/index.rst
```diff
@@ -6,23 +6,44 @@
...
 .. toctree::
    :maxdepth: 2 
-   :caption: Contents:
 
-   proxy-tutorial
-   rest-api
-   dev-overview
-   jsclient/index
+   introduction/index
+   specifications/index
```
loredanacirstea: As I said above, these should be moved to the `pyserver` docs. I proposed an order above.
Specifications should not be first. Installation & tutorials should be the first thing that a user sees.

docs/index.rst
```diff
@@ -6,23 +6,44 @@
...
+   applications/index
```
loredanacirstea: Applications is a bit confusing. It makes me think of applications built on top of uRaiden. 
How about `uRaiden Components` or `uRaiden Packages`, `uRaiden Installable Components`?
This should be right after the `Introduction`.

I would have actually thought of something like this (no Applications folder, which makes it harder to understand the docs imho):
```
introduction
tutorials
pyserver
pyclient
jsclient
contract
```

docs/tutorials/dev-setup.rst
```diff
@@ -0,0 +1,44 @@
+Development setup
```
loredanacirstea: This is duplicated. Installation instructions also exist in the pyserver docs. Maybe have only one place to explain installation and reference that. Just wanted you to have this in mind, doesn't need to be changed right now.

docs/tutorials/index.rst
```diff
@@ -0,0 +1,11 @@
+Tutorials
+=======================================
+
+.. toctree::
+   :maxdepth: 1
+
+   dev-setup
+   blockchain     
+   first-transaction
```
loredanacirstea: Consider using "First M2M microtransaction" as a name for this

docs/tutorials/index.rst
```diff
@@ -0,0 +1,11 @@
...
+   proxy-tutorial
```
loredanacirstea: Consider using "Paywall tutorial" as a name. "Proxy tutorial" doesn't say anything to the user.

docs/applications/pyserver/index.rst
```diff
@@ -0,0 +1,16 @@
...
+- a Python server that runs on the Receivers machine
+
+
+
+.. toctree::
+   :maxdepth: 2 
+
+   installation
+   usage
+   cmdline
```
loredanacirstea: This is empty when I build the docs - am I missing something? 
Btw, this is quite important for both pyserver and pyclient. Same for running the examples, which should also be documented.

docs/api/modules.rst
```diff
@@ -0,0 +1,7 @@
+microraiden
+===========
+
+.. toctree::
+   :maxdepth: 4
+
+   microraiden
```
loredanacirstea: Is there a current issue with the build or am I missing something? (there is no reference to the api in the docs and are not built correctly)

docs/tutorials/proxy-tutorial.rst
```diff
@@ -2,7 +2,12 @@ Proxy tutorial
 ########################################
 
 .. toctree::
-  :maxdepth: 2 
+  :maxdepth: 2
+
+Bevor starting, please follow the
```
loredanacirstea: Before

docs/tutorials/index.rst
loredanacirstea: Also, not part of the current changes, but this comment in the proxy tutorial is very important and should be put somewhere much more visible:
```
Off-chain transactions are stored in a sqlite database. You should do regular backups of this file – it contains balance signatures of the client, and if you lose them, you will have no way of proving that the client is settling the channel using less funds than he has actually paid to the proxy.
```

docs/index.rst
```diff
@@ -6,23 +6,44 @@
...
    contract/index
```
loredanacirstea: Because I cannot comment on the diagram, this should be removed: `docs/contract/diagrams/ContractClass.png`

loredanacirstea (commented):
Content wise looks good. 
I have some structure comments / questions though.

docs/applications/pyserver/index.rst
eorituz: You're right. It's missing for some reason.
Fixed with: https://github.com/raiden-network/microraiden/commit/ba1c77f65ffd4f179116a551ed9b3f92a85ff6ce

docs/index.rst
ezdac: we had this before and can roll it back

docs/applications/pyserver/index.rst
ezdac: This will auto-document the click options, have to look into it.
What's the status on that, wasn't that not supported at the moment?

docs/tutorials/proxy-tutorial.rst
ezdac: classic german mistake

docs/applications/pyserver/usage.rst
eorituz: afaik @agatsoh confirmed that the whole application worked.
@agatsoh can you confirm that Infura works in this case?

docs/index.rst
eorituz: I like bundling pyserver, pyclient and jsclient. They kind of belong together.

docs/applications/pyserver/index.rst
loredanacirstea: > What's the status on that, wasn't that not supported at the moment?
What are you referring to?

Also, instructions for building the docs would also be nice. Maybe I missed something. I just did a `sphinx-build -b html docs build`

docs/tutorials/dev-setup.rst
eorituz: You're right, the Idea was to move all installation and setup instructions to the `Development setup` section and reference it if needed.
I'm aware of that problem and will fix it in the next version.

docs/tutorials/index.rst
eorituz: I like the tutorial headline `Your first microtransaction`. It is the easiest and most straight forward tutorial and the name `Your first microtransaction` nicely encourages to try this tutorial first.

docs/tutorials/index.rst
eorituz: Fixed in: https://github.com/raiden-network/microraiden/commit/1d95ca2ee50c62bfc0efc17791dcd8e5eafb424a

docs/applications/pyserver/index.rst
ezdac: In the other PR #417 that's still open (and somewhat redundant, sorry), you commented:

   Docs show:
   microraiden [OPTIONS] COMMAND [ARGS]... and
   microraiden start [OPTIONS]

   Not valid commands at this point.

You probably mean that microraiden doesn't have a `scripts` entrypoint registered during setup?

The above `cmdline` docs just use a click auto-documenting plugin, so I don't know how much customization is available.
For now we can just document and explai the args passed to the example scripts manually (`--private-key` etc). This has some overlap with the click auto-documentation though.

docs/api/modules.rst
ezdac: Fixed in #436, that was supposed to be merged before any docs merge to master

loredanacirstea: So #436 is not ready yet I assume. Ok.

docs/applications/pyserver/index.rst
eorituz: The files `Installation.rst` and `usage.rst` are fine in my build. 
But in the file `../applications/pysercer/cmdline.rst` the section
```
.. click:: microraiden.click_helpers:main
   :prog: microraiden
   :show-nested:
```
doesn't compile correctly.

zum beispiel

@ezdac could you help me out with this issue?

loredanacirstea (approved):
My comments are mostly preferences to be considered. I am approving the PR. We can discuss docs structure later if needed.

docs/api/modules.rst
eorituz: @ezdac Could you explain why we have the `docs/api/..` folder?
Btw it compiles fine for me. @loredanacirstea can you access the files by manually entering the URL?

docs/tutorials/proxy-tutorial.rst
eorituz: Fixed: https://github.com/raiden-network/microraiden/commit/72d2d246bc79aad505ba37c41980aeae08e9a333

docs/api/modules.rst
loredanacirstea: yes, I can access them, but I only saw headlines with no other content.

PR #439, opened 2018-02-28
Docs: New design with modified Boostrap Template

ezdac:
We (@seichris mostly) created a version of the docs that modify the [sphinx-bootsrap-theme](https://github.com/ryan-roemer/sphinx-bootstrap-theme) in order to look similar to the other raiden pages.


ezdac:
Sorry @seichris, I did a squash commit (rewriting history to make one big commit out of several) because there was too much visual noise, so now no fame for you anymore :(

PR #443, opened 2018-03-05
update coincurve dependency

ofek:
https://github.com/ofek/coincurve#710


loredanacirstea:
@ofek , we would be happy to merge the PR if you rebase. We had to do a small version fix ourselves in the meantime, sorry for that.

ofek:
Done!

PR #452, opened 2018-03-07
Uninstall filter after get_logs() call

pcppcp:


loredanacirstea:
@pcppcp not required, but it would be nice to also add the `uninstall` function after this line: https://github.com/raiden-network/microraiden/blob/master/contracts/utils/logs.py#L146

Maybe after we settle on some libraries, we can use the same log handlers globally.

pcppcp:
added. You still need to call it manually. 
For deploy it doesn't really matter, but we should put this code i to a library common for both proxy and contracts deploy.

Issue #438, opened 2018-02-27
Make setuptools install routine work on ReadTheDocs


ezdac:
`python setup.py install --force` fails to install microraiden into a virtualenv.
This is crucial for having autodoc features in the deployed version of the documentation on http://mircoraiden.readthedocs.io .
The build inside a virtual environment there fails due to a depend
Unfortunately, this is how RTD installs projects:

```
update_docs_from_vcs(version)
if exists('setup.py'):
    run('python setup.py install')
if project.requirements_file:
    run('pip install -r %s' % project.requirements_file)
build_docs(version=version)
copy_files(artifact_dir)
```

So we have to rely on the setuptools install.
It seems that there is a problem of finding the `flask` package on the index?

I cannot confirm locally, because there ia different error occurs for `python setup.py install --force` on Mac OS.

Below is the output on RTD:

```
Installed /home/docs/checkouts/readthedocs.org/user_builds/microraiden/envs/docs-develop/lib/python3.5/site-packages/Flask_RESTful-0.3.6-py3.5.egg
Searching for flask_httpauth
Reading https://pypi.python.org/simple/flask_httpauth/
Downloading https://pypi.python.org/packages/6f/73/2a8da636f18ed580507a34483cb4fe26ef8d466909972029ebb3605415d2/Flask-HTTPAuth-3.2.3.tar.gz#md5=20f31ed9eb28ca04de57bece707c63d7
Best match: Flask-HTTPAuth 3.2.3
Processing Flask-HTTPAuth-3.2.3.tar.gz
Writing /tmp/easy_install-uwrbax4y/Flask-HTTPAuth-3.2.3/setup.cfg
Running Flask-HTTPAuth-3.2.3/setup.py -q bdist_egg --dist-dir /tmp/easy_install-uwrbax4y/Flask-HTTPAuth-3.2.3/egg-dist-tmp-gtizp2i_
creating /home/docs/checkouts/readthedocs.org/user_builds/microraiden/envs/docs-develop/lib/python3.5/site-packages/Flask_HTTPAuth-3.2.3-py3.5.egg
Extracting Flask_HTTPAuth-3.2.3-py3.5.egg to /home/docs/checkouts/readthedocs.org/user_builds/microraiden/envs/docs-develop/lib/python3.5/site-packages
Adding Flask-HTTPAuth 3.2.3 to easy-install.pth file

[...]

PR #454, opened 2018-03-08
Dependencies fix: pycryptodome 3.5.0 build failure; flask/setuptools bug

ezdac:
This will keep the used version of pycryptodome lower than 3.5.0, which encounters build failure (#455 ):
(https://github.com/Legrandin/pycryptodome/issues/142)

It will also make the non-pip setup routine with setuptools work (#438 ):
`python setup.py install`


ezdac:
There was a new version of `pycryptodome` released (`3.5.1`) that tackles the build error.
Now everything is working again without a version restirction.
I dropped the commit from this PR.

PR #456, opened 2018-03-09
Internal document links

agatsoh:
Links documents internally at suggested places


docs/applications/pyserver/installation.rst
```diff
@@ -1,60 +1,6 @@
 Installation
 ---------------
 
-It is required that you have python's `pip` installed.
```
ezdac: I wouldn't delete all this! It provides more info than the dev-setup tutorial 
(especially the non-dev setup with pip)

docs/applications/pyserver/installation.rst
```diff
@@ -1,60 +1,6 @@
...
-Using a global ``pip3`` installation
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-This method is not recommended, as it 
-can interfere with your global python installation.
-If you absolutely must, it still works though:
-
-.. code:: bash
-
-    sudo pip3 install microraiden
+Please go through the developer setup tutorial to set up all :ref:`requirements <dev-setup>`.
```
ezdac: You don't have to do this. The instructions above are enough theoretically.

Maybe we should merge this all with the dev-setup, and not call it dev setup anymore.
We should clearly separate developing on/for µRaiden (dev-setup) from using µraiden as a library (pip pypi mirror setup)
It should become clear though, that this is somewhat decoupled from the Javascript library.

docs/applications/pyserver/usage.rst
```diff
@@ -15,12 +15,13 @@ or
 
     python3 -m microraiden.examples.wikipaydia --private-key <private_key_file> --private-key-password-file <password_file> start
 
-By default, the web server listens on ``0.0.0.0:5000``. The private key
-file should be in the JSON format produced by Geth/Parity and must be
-readable and writable only by the owner to be accepted (``-rw-------``).
-A ``--private-key-password-file`` option can be specified, containing
-the password for the private key in the first line of the file. If it's
-not provided, the password will be prompted interactively. An Ethereum
-node RPC interface is expected to respond on http://localhost:8545.
```
ezdac: I think this belongs here, instead of the installation.
You don't need the node to install but to use it.

docs/specifications/http-headers.rst
```diff
@@ -1,3 +1,5 @@
+.. _http-headers:
```
ezdac: please use the `:doc:` directive (see other comment)

docs/tutorials/blockchain.rst
```diff
@@ -1,3 +1,5 @@
+.. _blockchain-setup:
```
ezdac: please use the `:doc:` directive (see other comment)

docs/tutorials/dev-setup.rst
```diff
@@ -1,3 +1,5 @@
+.. _dev-setup:
```
ezdac: please use the `:doc:` directive (see other comment)

docs/tutorials/first-transaction.rst
```diff
@@ -1,3 +1,5 @@
+.. _first-transaction:
```
ezdac: please use the `:doc:` directive (see other comment)

docs/applications/pyserver/components-overview.rst
```diff
@@ -1,6 +1,12 @@
+
 Components overview
 --------------------
 
+This section will give you an overview of the classes used to build the uRaiden proxy. It should help keeping track of functionality of the proxy server.
+The Illustrations are inspired by UML even though they might not be exact standard.
+There is an overview for the Channel Manager class
```
ezdac: should this sentence be in there?

docs/applications/pyclient/components-overview.rst
```diff
@@ -1,3 +1,5 @@
+.. _client-comp-overview:
```
ezdac: if you **don't** want to link to a specific section in a doc, I would rather not define unique refs and then link to them.
To link to whole documents directly, please use the `:doc:` directive (as explained [here](http://www.sphinx-doc.org/en/stable/markup/inline.html#role-doc) )

docs/applications/pyserver/installation.rst
eorituz: Well you're right max. Playing everything in the Dev Setup is most likely the best way to solve it.
We could rename it to "installtion" and explain what "pip install microraiden" does. (I'm not really sure what you can do with this installation)

ezdac:
please rebase against `docs-develop` again, had to include the latest changes from master - sorry about that

docs/applications/pyserver/installation.rst
ezdac: I would say that's actually the main installation process people will (should) use, once we have a stable release and declare the API fixed.

You then can use µRaiden as a library and inherit your custom classes (e.g. handling some specific resource) from the `Expensive` [class](https://github.com/raiden-network/microraiden/blob/master/microraiden/proxy/resources/expensive.py#L23) etc.

docs/applications/pyserver/components-overview.rst
agatsoh: We can leave it there for the time being.

ezdac: This is obsolete since the new updates from #462

ezdac:
closed in favour of follow up #463

PR #479, opened 2018-07-12
Update requirements-dev.txt

krisnaer:
added requirements that were found to be needed when doing a fresh install of µRaiden


ezdac:
Those are not python dependencies but system dependencies.
They can't just be appended to a dependencies files, but have to be installed manually.
We should add system wide dependencies in the documentation though.

ezdac:
Follow up issue #486

PR #481, opened 2018-07-31
Return proper json from /channels resource

pcppcp:


pcppcp:
damn the contract test are in a very bad shape when it comes to `flake8`. I've disabled the most annoying checks and fixed everything else I could. 

Related: #482

loredanacirstea:
Hmmm, yes - I did not fix what did not bother me at that point. Let me know what I can help with to clean up my mess.

